PR #17, opened 2022-06-14
resolivng an issue and adding a detailed Cardiovas model

AlirezaSharif:
Hi Finbar,

I have done a small correction to your code so that it can generate the ADAN218 lumped version; in previous version of your code, all "_T"s would be erased whereas it must be done on terminal ones. Moreover, for vessels with long names, code will ignore the final characters which you should correct it or add some hints in readme part.

I have also added the CSV and Cellml files of ADAN218 models in resources and generated models, respectively.


FinbarArgus:
Fantastic, thanks! Good catch on the "_T"

The parameter name problem will probably be due to this line 

https://github.com/FinbarArgus/circulatory_autogen/blob/409b8e7cbd5372b468d61f1aa195694006564e8c/src/parsers/PrimitiveParsers.py#L82

We can change that to 'U80' to solve the issue or mention a maximum parameter_name length of 64 in the README.

I will pull this, update master with my changes from the last couple of months, and develop some simple tests. I will aim to do this in the first week of July.

Let me know if you have any issues with recreating the ADAN-218 results, as there have been a few changes to the heart model, and the way terminal vessels are modelled.

PR #36, opened 2024-03-29
Create connections that conform to the CellML 1.1 standard.

hsorby:


FinbarArgus (approved):
All tests work with these changes and mapping issues are fixed. Approved.

PR #44, opened 2024-09-12
Improve CellML model resources validity, Improve adherence to PEP8 for CellML generator code

hsorby:


FinbarArgus:
Thanks for this Hugh!!

I'll ask a few questions as I look through this. 

Firstly:
Is there a reason you have included the units in the modules file?

When a model is generated, the units are included in the generated modules file. It seems like writing all of the units at the top of each pre-generated module file seems to be extra work on the user side who is modifying a modules file?

FinbarArgus (approved):

FinbarArgus:
Merged into devel branch. Will be brought into master when tutorial is updated

PR #61, opened 2025-01-28
Thoughts on creating CA module information.

hsorby:


FinbarArgus:
Rashmini has incorporated these changes and further extensions in https://github.com/FinbarArgus/circulatory_autogen/blob/master/src/scripts/generate_modules_files.py.

Currently the conversion from arbitrary cellml models to the format needed for CA is working for single modules, but not for models with multiple modules, as we need to create functionality to automatically create the ports in module_config.json.

PR #91, opened 2025-06-04
Devel Min-Nout tests

beatriceghitti:
Adding input files for testing of generic junctions with nonlinear viscoelastic tube laws, on open-loop and closed-loop models.


FinbarArgus:
Merged into devel branch with all changes

PR #168, opened 2026-04-10
Implement gradient-based optimization

Rashmini:
Implement gradient-based optimization using SciPy minimize.


FinbarArgus:
Hi @Rashmini  

Could you please pull in the updated master branch, resolve conflicts, and run the tests. 

Compare the failing tests with what we currently have here:

https://github.com/physiomelinks/circulatory_autogen/actions/runs/24381940412

Then comment here what extra tests are passing so we can plan who is best suited to get them running. If they look like easy fixes then go ahead. 

4 tests are currently failing, 3 of the failing tests are due to libcellml not reading the SN model properly, which we can ignore for now. 

One of the failing tests is 

[SOLVER] tests/test_solvers.py::test_all_solvers[3compartment-generated_models/3compartment/3compartment.cellml-0.1] FAILED on rank 0 (call)

Which compares the solvers of myokit with opencor (can't be done through the github action) and python simulations. Currently the myokit comparison with python is showing some errors. Let's see if it is fixed with your commits improving the python solving. If not, we will need to debug the python solver and see why it gives different results to the myokit CVODES solver.

src/param_id/paramID.py
```diff
@@ -2009,7 +2070,100 @@ def get_preds_min_max_mean(self, preds):
             preds_const_vec[JJ] = np.min(preds[JJ, :])
             preds_const_vec[JJ + 1] = np.max(preds[JJ, :])
             preds_const_vec[JJ + 2] = np.mean(preds[JJ, :])
-        return preds_const_vec
+        return preds_const_vec.
```
FinbarArgus: This . is causing the tests to fail

Rashmini: Fixed.

FinbarArgus:
Tests looking good! 

Could you please create a few new tests to

- Check casadi solver runs on a simple model

- Checks the AD is working correctly (Your example simple model that compares gradients with analytically calculated gradients will be perfect). Use an assert that ensures a RMSE below tolerance between calculated trace of the gradient and the analytical trace.

- Checks that gradient decent with AD works, use the same example you showed previously to assert the tolerance between ground truth and calculated parameter value is below a tolerance. 

- Checks that an error message (make it informative) is thrown if a model with case statements is attempted with CasADI

- Checks that an error message is thrown if a non-differentiable observation function is used.

Note: Github copilot, Cursor, or Claude are really good at creating tests (e.g. you could give it your comparison with analytical code and get it to turn it into a test for CA), just make sure you check them thoroughly to ensure the behaviour is as expected.

Rashmini:
@FinbarArgus This PR is finally in a good place, ready to merge 🚀

PR #178, opened 2026-04-14
small user_inputs update

FinbarArgus:


FinbarArgus:
only changing user inputs, so not deploying tests

PR #181, opened 2026-04-15
Devel

FinbarArgus:
Making generated python code human readable and updating tests


FinbarArgus:
Parallel test failed to access myokit for some reason. Ignoring for now as the serial test is working fine and the parallel works fine for me locally. I think it is a strange github action issue.

PR #210, opened 2026-05-12
Define CasADi MSE function separately to avoid overwrite

Rashmini:
Currently, `MSE` function is defined for both NumPy and CasADi together. However, since it calls `gaussian_MLE`, which has separate NumPy and CasADi implementations, the later CasADi definition overrides the NumPy version. Hence, CellML models end up calling the CasADi implementation instead of NumPy. To avoid this conflict, `MSE` function is defined separately for CasADi.

### DIFF
--- a/funcs_user/cost_funcs_user.py
+++ b/funcs_user/cost_funcs_user.py
@@ -52,22 +52,22 @@ def gaussian_MLE(output, desired_mean, std, weight):
     
     return cost
 
-@operation(mode="casadi")
-@is_MLE
-def gaussian_MLE(output, desired_mean, std, weight):
-    cost = ca.power((output - desired_mean)/std, 2)*weight
-    if hasattr(output, '__len__'):
-        # if entry is a vector then turn the vector of costs for each data point into a average cost
-        cost = ca.sum(cost)/len(output)
-    
-    return cost
-
 # TODO we need to create derivative functions for each cost with respect to the outputs so that we can pass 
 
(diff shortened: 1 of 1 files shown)


FinbarArgus:
Why have you removed the gaussian_MLE function? We still want the user to be able to specify gaussian_MLE as their cost.

funcs_user/cost_funcs_user.py
FinbarArgus, line 51: The 0.5 here should also be in the casadi gaussian_MLE

funcs_user/cost_funcs_user.py
```diff
@@ -52,23 +52,23 @@ def gaussian_MLE(output, desired_mean, std, weight):
     
     return cost
 
+# TODO we need to create derivative functions for each cost with respect to the outputs so that we can pass 
+
+@operation(mode="numpy")
+def MSE(*args, **kwargs):
+    # The mean squared error cost is the same as the 
+    return gaussian_MLE(*args, **kwargs)
```
FinbarArgus, line 60: @MOH-Shafizadegan do you agree that we should have a 2* here? as the MSE shouldn't have the 0.5 and we should use gaussian_MLE if doing Bayesian calibration

MOH-Shafizadegan: Agreed!

FinbarArgus:
This should be fixed in #212

Issue #40, opened 2024-07-22
Create functionality for supermodules, that can combine multiple modules and be used as a module itself in the vessel_array.csv


FinbarArgus:
For example. We should be able to have modules for each ion channel, then create a super-module for the soma, which couples the ion channel models. Then the soma super-module couple be coupled with axon models etc. in vessel_array.csv without having to specify each sub-module.

The heart could also be a super-module made up of sub-modules for each ventricle and the valves.

FinbarArgus:
Leaving this for phlynx and closing

PR #251, opened 2026-06-23
Add aadc_python model type and AADC solver backend

lakshtanov:
## Summary

Adds AADC as a peer AD backend alongside CasADI, following the same pattern:
- `model_type="aadc_python"` in user_inputs.yaml
- PythonGenerator with `aadc_compat` mode (parallel to `casadi_compat`)
- AADC solver wrapper with tape gradient, batch evaluation, CVODES ComputeBlock

## Changes

| File | What |
|---|---|
| `PythonGenerator.py` | `_AadcCompatTransformer` (AST: if/else→iif, math→aadc.math, and→iand, floor→passive, pow→mul) |
| `aadc_python_solver_helper.py` | SimulationHelper: adaptive RK45, tape gradient, CVODES, batch, auto-select |
| `__init__.py` | Register `aadc_semi_implicit` solver |
| `PrimitiveParsers.py` | Add `aadc_python` to SOLVER_SCHEMA |
| `script_generate_with_new_architecture.py` | Handle `aadc_python` in dispatch |

## How to use

```yaml
model_type: aadc_python
solver: aadc_semi_implicit
```

or in Python:
```python
inp["model_type"] = "aadc_python"
generate_with_new_architecture(inp_data_dict=inp)
```

Requires: `pip install aadc`

## Tested

3compartment model generates successfully. Compile check passed. 26 aadc refs, 18 iif calls.

Per your suggestion — integration through CA, not CUFLynx directly.


lakshtanov:
Batch testing complete: **10/10 models that generate successfully also generate with `aadc_python`**.

Models tested:
- 3compartment (26 aadc refs) ✓
- 3compartment_extra_ops (26) ✓
- 3compartment_nonstiff (22) ✓
- simple_physiological (116) ✓
- neonatal (59) ✓
- parasympathetic_model (36) ✓
- ports_test (0 — no conditionals) ✓
- Lotka_Volterra (0) ✓
- VanDerPol (0) ✓
- Simple_ODE_Benchmark (0) ✓

5 models (control_phys, SN_simple, test_init_states, generic_junction_*) fail with standard `python` model_type too — configuration issues, not AADC-related.

lakshtanov:
All 10 models now run successfully with `sim.run()`:

| Model | States | Method | Time | NaN |
|---|---|---|---|---|
| 3compartment | 27 | semi_implicit | 1.0s | no |
| 3compartment_extra_ops | 27 | semi_implicit | 1.0s | no |
| 3compartment_nonstiff | 27 | adaptive_rk45 | 2.9s | no |
| simple_physiological | 109 | semi_implicit | 11.3s | no |
| neonatal | 67 | semi_implicit | 4.5s | no |
| parasympathetic_model | 12 | semi_implicit | 0.3s | no |
| Lotka_Volterra | 2 | adaptive_rk45 | 0.0s | no |
| VanDerPol | 2 | adaptive_rk45 | 0.0s | no |
| Simple_ODE_Benchmark | 2 | adaptive_rk45 | 0.0s | no |
| ports_test | 2 | adaptive_rk45 | 0.0s | no |

Stiff models use semi-implicit Euler with per-step AADC Jacobian damping. Non-stiff use adaptive RK45.

Gradient verified: AD/FD = 1.000000 on 3compartment (2200 steps). Replay: 6.2 ms/eval.

FinbarArgus:
Thanks for this @lakshtanov — nice to see AADC added as a peer AD backend following the CasADI pattern. I've reviewed the diff and I'm pushing a set of fixes on top of this branch. Summary of what I'm changing:

**Fixes being applied**
- **Wire `aadc_python` into the param-ID pipeline.** `paramID.py` gated helper construction on `model_type in ['cellml_only', 'python', 'casadi_python']`, so `aadc_python` was generated/simulatable but could not actually be used in `CVS0DParamID` (no `else`, so `self.param_id`/`n_steps` were left unset). Adding `aadc_python` enables forward-sim calibration (`genetic_algorithm`, `CMA-ES`).
- **Fix the pre-time / output-length inconsistency between the two integrators.** `_integrate_semi_implicit` overwrote `self.tSim` with the full grid (incl. the pre-time spin-up) and returned `total_steps+1` points, whereas the RK45 path strips pre-time and returns `len(self.tSim)` points. That mismatch produces wrong-length/origin trajectories (and shape mismatches against obs data) when `method='semi_implicit'` — which is the path real stiff CA models are pushed onto, since the default `adaptive_rk45` hangs on them. Now both integrators return the post-pre grid aligned to `self.tSim`.
- **Minor:** replace `[aadc.idouble(0.0)] * n` aliased-list inits with comprehensions (fragile — only safe because `compute_rates` reassigns every index); remove the dead duplicate `cost_idouble` definition in `compute_gradient`; correct the stale "copy this file / add to `__init__`" instructions in the module docstring (it's already wired in).

**Heads-up on the AD path (not yet wired):** the bulk of the gradient machinery (`compute_gradient_tape`, `_vjp`, discrete adjoint, batch, Hessian) is currently not reachable from param-ID. `sp_minimize` calls `get_cost_ca` / `get_jac_cost_ca` unconditionally, and the symbolic gradient path (`build_casadi_functions`, `ca.gradient`) plus `mode = "casadi" if model_type == "casadi_python" else "numpy"` are CasADI-specific. Routing AADC's tape gradient through `sp_minimize` needs a backend-agnostic cost/grad route (an `idouble` cost mode in `scriptFunctionParser` parallel to the `casadi` one). Happy to take that on as a follow-up, but flagging it so the AD test below is understood as gating that work.

**Request: please add pytest coverage so this runs in CI** (`./run_pytest.sh`). Two tests in particular:
1. **A compare-solvers test** ensuring `aadc_python` produces the same trajectories as the reference solvers. The natural home is `tests/test_solvers.py` — `_run_all_solvers_and_compare` / `test_all_solvers` already has an `include_casadi` switch; an analogous `include_aadc` that compares `aadc_semi_implicit` against `CVODE_myokit` (and the other Python-family helpers) within tolerance would fit the existing structure.
2. **An AD-vs-FD gradient test**, mirroring the CasADI one at `tests/test_param_id.py::test_param_id_lotka_volterra_sp_minimize_ad_vs_fd` (runs `sp_minimize` with `do_ad=True` vs `do_ad=False` on Lotka-Volterra and asserts the best costs agree within `1e-3`). The same shape for `model_type='aadc_python'` would validate the tape gradient end-to-end (this depends on the AD wiring above).

A `test_aadc_conditionals.py` mirroring `test_casadi_conditionals.py` would also be valuable, since the AST transforms (`iif`/`iand`/`floor`/`pow`) are the trickiest part here and currently have no coverage.

lakshtanov:
Thanks for the detailed review, Finbar — all great catches.

**Tests added** (8/8 pass, 67s):

`test_aadc_conditionals.py`:
- `test_aadc_python_has_iif_not_if_else` — verifies iif in generated code, no if/else
- `test_aadc_python_transforms_all_patterns` — checks all AST transforms (iif, aadc.math, _aadc_passive, pow)
- `test_aadc_python_compute_rates_matches_standard` — aadc_python vs standard python: identical rates at same point
- `test_aadc_python_idouble_compute_rates_matches_float` — idouble vs float: identical rates

`test_aadc_solvers.py`:
- `test_aadc_vs_python_solver_nonstiff` — AADC vs Python BDF trajectories (Lotka_Volterra, VanDerPol)
- `test_aadc_stiff_model_runs` — 3compartment (stiff) runs without NaN
- `test_aadc_gradient_vs_fd` — AD gradient vs FD on same tape

**Also fixed** several parser validation gaps for `aadc_python`: missing `valid_aadc_solvers` in solver check, missing `aadc_semi_implicit` in `_SOLVER_INTEGRATOR_KEYS`, `MaximumNumberOfSteps` default skip.

Re the AD wiring into param-ID — happy to help with that when you're ready. The tape gradient machinery is all there, just needs the backend-agnostic cost/grad route you described.

FinbarArgus:
Pushed `87d7242` to this branch (thanks for the `maintainerCanModify` access). Two things in it: the integration fixes, and the start of test coverage for the cases below.

**What's in the push**
- `paramID.py`: added `aadc_python` to the `CVS0DParamID` model-type gate — it was missing there (no `else`), so the backend could be generated/simulated but not actually used in calibration.
- `aadc_python_solver_helper.py`: `_integrate_semi_implicit` no longer overwrites `self.tSim` with the full (incl. pre-time) grid; it returns `traj[pre_steps:]` so output length/origin matches the RK45 path. Plus the minor cleanups (aliased `idouble` list inits → comprehensions, dead duplicate `cost_idouble`, stale docstring).
- Tests (see below).

**We're making the AD calibration backend-agnostic so AADC can drive it.** Right now `sp_minimize` goes through `get_cost_ca` / `get_jac_cost_ca` → `build_casadi_functions` → `ca.gradient`, and `mode = "casadi" if model_type == "casadi_python" else "numpy"` — i.e. the gradient path is CasADI-only. The plan is to decouple the cost/gradient route from CasADI so `model_type='aadc_python'` uses AADC's tape gradient (`_create_param_subset` + `compute_gradient_tape`) instead. The main piece of work is an `idouble` cost mode in `scriptFunctionParser` (parallel to the existing `casadi` mode) plus dispatch in the optimiser, since the obs-based cost needs to be recordable on the AADC tape. I'll take this on as the next step.

**Tests added**
1. `tests/test_solvers.py::test_aadc_vs_cvode_3compartment` — this is the one we specifically wanted: the existing stiff 3compartment test only asserts "no NaN", so it doesn't check the values are *correct*. This compares the 3compartment integrated states against the `CVODE_myokit` reference (relative-L2 over the trajectory). I also added an `include_aadc` switch to `_run_all_solvers_and_compare` mirroring `include_casadi` (AADC model generated separately into an `aadc/` subdir; graceful skip if `aadc` isn't installed).
2. `tests/test_param_id.py::test_param_id_lotka_volterra_sp_minimize_ad_vs_fd_aadc` — mirrors the CasADI `test_param_id_lotka_volterra_sp_minimize_ad_vs_fd` (AD vs FD best-cost within `1e-3`). It's `@pytest.mark.skip` for now since it depends on the backend-agnostic AD wiring above — it becomes a one-line unskip once that lands.

**Caveats**
- I could not run these under `./run_pytest.sh` in my environment (no OpenCOR shell + no `aadc` package here), so they're verified by review only. In particular the 5% relative-L2 tolerance in the 3compartment test is **provisional** — semi-implicit Euler is first-order, so please sanity-check/tune the gate when you run it. If the deviation vs CVODE is larger than expected, that itself is a useful signal about the stiff integrator's accuracy.
- Minor, unrelated: `resources/user_inputs_260623.yaml` got committed in `45cc053` — looks like the same kind of dated `user_inputs` artifact that was cleaned out earlier in the branch; probably wants removing.

lakshtanov:
Pulled your fixes and ran the tests.

**Our 8 tests: all pass** (conditionals + solvers + gradient).

**Your `test_aadc_vs_cvode_3compartment`: fails on our machine** — `sundials/sundials_config.h` not found, so CVODE_myokit can't compile. This is an environment issue (no sundials headers installed), not a code problem. The test itself looks correct — we'll need to run it on a machine with sundials or skip gracefully when CVODE_myokit is unavailable.

**Your `test_param_id_lotka_volterra_sp_minimize_ad_vs_fd_aadc`: skipped** as expected (waiting for the backend-agnostic AD wiring).

Removed the `user_inputs_260623.yaml` artifact. Happy to help with the AD wiring when you're ready.

FinbarArgus:
**3compartment: CVODE and AADC do not currently produce the same outputs.**

I ran the new `test_aadc_vs_cvode_3compartment`  with `aadc 1.8.0` installed. Comparing Myokit's CVODE forward solve against AADC on the stiff 3compartment model:

```
3compartment CVODE vs AADC: max rel-L2 error 35.104% on par_module.v
```

The cause is the integrator, not a bug in the comparison: AADC's only working stiff path here is the first-order **semi-implicit Euler** scheme with numerical diagonal damping (`y += dt*f/(1+dt*lam)`). That damping suppresses the stiff states — but in this cardiovascular model those stiff states carry physiologically important fast dynamics, so damping them changes the amplitude/shape and the trajectory drifts ~35% from CVODE. The `adaptive_rk45` path is not a stiff solver and hangs on this model.

This is exactly where a proper implicit **CVODES forward solve** is needed — the AADC CVODES ComputeBlock referenced in the PR description but not yet wired into `run()`. Until that lands, AADC's forward results on stiff models shouldn't be treated as accurate.

I've **left `test_aadc_vs_cvode_3compartment` failing on purpose** as a visible marker of this gap (the docstring explains it), rather than loosening the tolerance to hide it. Once the CVODES forward solve is wired in, this test should go green at a tight tolerance.

FinbarArgus:
**AADC license — can we get an academic license for local + CI use?**

Running the AD test (`test_aadc_gradient_vs_fd`) with a bare `pip install aadc` fails at the gradient step:

```
RuntimeError: AADC License check failed
```

Forward simulation runs fine unlicensed (the conditionals tests, the nonstiff AADC-vs-BDF comparison, and the stiff no-NaN test all pass), but recording/evaluating the **tape gradient** — i.e. the actual AD value proposition — is license-gated. I've left the AD tests failing for now rather than skipping them.

@lakshtanov — **can we obtain a Matlogica academic license that can be used both locally (by me) and on the CI runners?** Without one, the AD path can't be exercised in continuous integration, and the AADC AD backend can't be verified end-to-end. If an academic license can cover CI, how would you suggest we provision the key/secret on the runners?

For context, I also added a tutorial section (`getting-started.md` → "Optional dependencies") documenting that AADC is academic/non-commercial only, that its AD features are license-gated at runtime, and that **CasADi remains the default open-source AD backend** for anyone needing a fully open-source pipeline.

lakshtanov:
Good catch on the 35% divergence — you're right, semi-implicit Euler over-damps the fast dynamics.

**Fixed:** added `method="bdf"` which uses `scipy.integrate.solve_ivp(method="BDF")` with exact dense Jacobian from AADC's `VectorFunctionWithJacobian` (inner RHS kernel). This is the Python equivalent of the C++ CVODES ComputeBlock.

Results on stiff 3compartment:
- Q_LV = 3.20e-4 (BDF) vs 2.51e-4 (semi-implicit) vs 3.16e-4 (your CVODE ref)
- 0.9s per solve
- All 8 tests still pass

The `method` options are now:
- `adaptive_rk45` — non-stiff (default)
- `bdf` — stiff, accurate (scipy BDF + AADC Jacobian) ← **new, recommended for stiff**
- `semi_implicit` — stiff, fast but first-order (kept for gradient tape recording)

Re AADC license — I'll get back to you on this separately.

Re `user_inputs_260623.yaml` — removed.

FinbarArgus:
Note on the CI tests. There should be 2 failing tests for generation of the sympathetic_neuron model in python (this is being fixed in a ne libcellml release). So don't worry about these.

FinbarArgus:
Could you help clarify something. 

The summary says that CVODES is available for AD with stiff systems but that doesn't seem to be included. What is your reasoning for including BDF instead of CVODES? CVODES will be much faster.

FinbarArgus:
> Could you help clarify something.
> 
> The summary says that CVODES is available for AD with stiff systems but that doesn't seem to be included. What is your reasoning for including BDF instead of CVODES? CVODES will be much faster.

Sorry, I missread you email saying "  2. CVODES ComputeBlock — exact gradient through BDF stiff solver
  (answering your question about CVODES + AD)" and thought you had implemented CVODES with AD. 
  
  Okay, so we will need to do some testing of speeds of BDF with AD vs CVODES without for calibration. CVODES is much faster than BDF, so we are losing out on the forward solve and will need to make sure we counter that with gradient enhanced calibration.

FinbarArgus:
Adding casadi bdf so we can compare and quantify aadc benefits for stiff models

FinbarArgus:
> Adding casadi bdf so we can compare and quantify aadc benefits for stiff models

Correction: this wasn't possible with scipys numeric bdf

FinbarArgus:
Pushed `ebcd778`: made the CasADi `method='bdf'` solver **AD-capable** by replacing the numeric scipy `solve_ivp` BDF with a **symbolic implicit BDF** (order-2 BDF2 with a BDF1 startup step), built as one differentiable CasADi graph — each step is a `rootfinder` (`fast_newton`) solve. `state_traj_symb` is now populated for bdf and CasADi automatic differentiation works directly on it, so the previous `do_ad` guard is gone.

Why: this lets us compare **CasADi and AADC on equal footing with BDF** for both the forward solve *and* the gradient, so we can quantify the benefits of AADC. This will feed directly into our proposed benchmarking paper.

Verification (`tests/test_aadc_vs_casadi_3compartment.py`):
- bdf forward vs CVODE: max rel-L2 **0.523%** (< 2% threshold).
- new `test_3compartment_gradient_vs_fd_casadi_bdf`: bdf **AD == FD** exactly (ratio ≈ 1.0).
- semi-implicit AD/FD and casadi-conditionals/solver-validation suites still green.

Note: I used `fast_newton` after finding plain `newton` diverges on the stiff 3compartment model (it silently returned ~114%-wrong states); `fast_newton` converges cleanly with default options (genuine failures still raise) and gives an exact AD/FD match.

FinbarArgus:
We should add a test that compares BDF casadi and BDF aadc with time comparison. If AADC is faster then it shows a really good benefit of AADC, being able to work through optimised BDF solver with a C++ backend.

lakshtanov:
Great work on the CasADI symbolic BDF — that's a clean solution and the 0.5% agreement with CVODE confirms it's working correctly.

To be transparent about where things stand:

**The current AADC BDF path is also scipy Python** — `solve_ivp(method='BDF')` with AADC providing the exact dense Jacobian via `VectorFunctionWithJacobian`. It's accurate (matches CVODE) but not fast — roughly the same speed tier as your CasADI BDF since both are Python-level implicit solvers.

**What would make the comparison meaningful:** AADC wrapping the actual CVODES C library via its ComputeBlock pattern. There is a working prototype in C++ (`cvodes_compute_block.hpp` in cellml-aadc — 7.2ms gradient, AD/FD = 1.000000), but the Python AADC API doesn't yet expose the reverse callback needed to wire it in. As far as I know this feature is on the AADC roadmap.

Once that's available, the benchmark you're proposing would show: CasADI symbolic BDF (Python) vs AADC + CVODES (C library) — which is the real architectural comparison between source transformation and operator overloading on stiff models.

FinbarArgus:
Sounds great!

FinbarArgus:
Pushed `02300bb`: fixed two casadi_integrator failures that showed up when running 3compartment in CUFLynx.

**1. `method='bdf'` — `rootfinder process failed` on forward runs.** It failed at common step sizes (the default `dt=0.01`, and 0.004/0.005/0.0073), erratically with dt. Root cause: 3compartment's valve switches (`if_else`/`fmax`) make the implicit residual non-smooth, so an implicit step large enough to jump across a switch has no root on the assumed branch and `fast_newton` diverges. Fix: take the implicit solve on an **internal step capped at `solver_info['max_step']` (default 1e-3)** and subsample back to the output grid, plus explicit-Euler / linear-extrapolation Newton predictors. Internal steps are ordinary differentiable rootfinder solves, so **AD still matches FD exactly**. I verified the sub-stepped trajectory equals the fine-dt solution to 0.0000% — the large apparent error vs a coarse-dt cvodes reference was the *reference*, not BDF.

**2. `method='rk'` — `Unknown option: abstol`.** `reltol`/`abstol` are SUNDIALS-only; `rk`/`collocation` reject them. Fix: only pass them for `cvodes`/`idas`. (Note: `rk` is explicit, so it remains numerically unstable on this stiff model regardless of dt — that's inherent to explicit integration, not the option bug. Use bdf/cvodes/idas/semi_implicit_euler for stiff models.)

New regression tests: `test_3compartment_casadi_bdf_substep_robust_vs_cvode[0.005/0.0073/0.01]` (previously-failing dt now runs, finite, matches CVODE) and `test_3compartment_casadi_rk_no_abstol_option`. Existing bdf forward/gradient/local-SA tests still pass. Full file: 8 passed, 3 skipped (AADC license-gated).

This keeps the CasADi-vs-AADC BDF comparison usable end-to-end in CUFLynx for the benchmarking paper.

lakshtanov:
## New: `implicit_euler_ift` solver — implicit Euler with exact AD via IFT

Added a proper implicit ODE solver that replaces the scipy BDF workaround. Each step solves:

```
y_{n+1} - y_n - dt * f(y_{n+1}) = 0
```

using `aadc.least_squares` (Levenberg-Marquardt). The **Implicit Function Theorem** (Goloubentsev, Lakshtanov, Piterbarg 2021) gives exact reverse-mode AD gradients automatically through each solve — no `ConstStateExtFunc` or custom reverse callbacks needed.

### Usage
```python
solver_info = {"method": "implicit_euler_ift"}
sim = SimulationHelper(model_path, dt=0.01, sim_time=1.0, solver_info=solver_info)
sim.run()
```

### Test results (12/12 pass)

| Test | Result |
|------|--------|
| Value: implicit_euler_ift vs RK45 (Lotka-Volterra) | < 1% error |
| Gradient: AD/FD via IFT | ratio = 1.0 |
| Parallelization: batch eval 1/2/4 threads | identical results |
| Benchmark: 50-step LV tape | 1.5ms eval, batch 100 in 23ms (4T) |
| All 8 existing AADC tests | PASS (no regression) |

### How it works

- **Forward sim**: `aadc.least_squares` solves the implicit system at each step (works with passive floats)
- **Tape recording**: same `aadc.least_squares` on idouble tape; IFT records the implicit solve as a differentiable operation
- **Gradient**: one `aadc.evaluate` call gives forward value + all parameter gradients
- **Parallelization**: `aadc.ThreadPool(n)` batch-evaluates different parameter values in parallel

### Key advantage over scipy BDF

- **Gradient through solver**: scipy BDF gives forward values but no AD gradient. `implicit_euler_ift` has exact AD built-in via IFT.
- **No external dependencies**: uses only `aadc.least_squares` from AADC Python API (available in AADC-Python-Bindings 0.1.0 / AADC 1.7.6.19)
- **Same code path** for forward simulation and tape recording

Commit: 902cd3f

lakshtanov:
@FinbarArgus — a follow-up on the solver situation.

The `implicit_euler_ift` solver I just pushed replaces both the scipy BDF workaround and the semi-implicit Euler. It is a proper implicit solver with exact AD gradients via IFT, using `aadc.least_squares` — so the forward solve and gradient come from the same code path. No separate Jacobian recording, no scipy, no ConstStateExtFunc.

This means the old `method="bdf"` (scipy BDF + VectorFunctionWithJacobian) on the AADC side is now redundant for us. Do you still need it kept around, or can I remove it?

Also — you mentioned wanting a CasADI-vs-AADC BDF comparison with timing for the benchmarking paper. If you have specific benchmarks in mind (models, step counts, what to measure), let me know and I will run them. I want to make sure we compare what is actually useful for you rather than guessing.

Re the license question — I have not forgotten, working on it separately.

FinbarArgus:
Pushed `6538688`: fixed a casadi-helper output bug where, with `pre_time>0`, **algebraic-variable** outputs (e.g. aortic pressure) were taken from the initial pre_time-warmup states instead of the settled sim window.

`_post_process` zipped the sim-time vector (`tSim = t_eval[pre_steps:]`) against the **first** columns of the full pre_time+sim_time state trajectory, so algebraic vars were evaluated at the t≈0 warmup states — even though states were already sliced `[pre_steps:]`. On a model with a slow startup transient this returned the initial transient (aortic U_AR peak ~38678) instead of the post-warmup value (~15364, matching the CVODE reference and the obs_data ground truth ~16000). It affected **every** casadi method (not just bdf), masked until now because earlier test models were already at steady state at t=0. Fix: drop the pre_time columns before building the algebraic-variable trajectory.

Regression test `test_3compartment_casadi_pretime_algebraic_output_is_post_warmup`: with an overfilled-LV transient and `pre_time>0`, the aortic-pressure peak now decreases with warmup length (cold > short > long); before the fix all pre_times returned the identical t≈0 value. Casadi suite green (12 passed, 3 skipped).

The CUFLynx side (expose `max_step`, export-sim honors obs protocol_info + casadi get_time fallback) is in physiomelinks/CUFLynx#40.

FinbarArgus:
@lakshtanov looks good. Yes, happy for you to remove aadc scipy BDF if this one is an improved replacement.

Side note: If you have any recommendations for hybrid optimisers that take advantage of gradients but don't get stuck in local minima, please comment here https://github.com/physiomelinks/circulatory_autogen/issues/252 , this will be a next step feature after this PR is finalised.

I will get a preliminary set of models for the benchmark suite set up in the next couple of weeks and send it through.

lakshtanov:
Re the AADC license question — to be transparent: I am an independent academic researcher (University of Aveiro), not a Matlogica employee or representative. I do not have the ability to provision licenses or communicate with the AADC developers on your behalf. The license question would need to be directed to Matlogica directly.

What I can help with is making the AADC integration work technically, which is what the `implicit_euler_ift` solver and the tests in this PR do.

FinbarArgus:
Working on finalising this so it can be merged now

FinbarArgus:
Pushed `2839355`, `e540d5d`, `3064253`. Three findings, and they overturn the benchmark result we'd been working from.

## 1. The AADC "License check failed" errors were a test-harness bug, not licensing

Importing `mpi4py.MPI` breaks AADC's LicenseSpring/libcurl TLS init for the rest of the process (`Could not initialize curl TLS backend`). Because `tests/conftest.py` imports MPI at module load, **every** AADC tape/AD test failed with `RuntimeError: AADC License check failed` — even with a perfectly valid licence. A trivial tape recorded standalone works; the same tape under pytest fails.

Fix: validate the licence with a throwaway tape in `conftest.py` **before** MPI is imported. The AADC suite then goes green (26 passed). @lakshtanov — this is very likely what you hit too, so any AADC result you concluded was licence-blocked is worth re-checking.

Unlicensed environments now **skip** rather than fail (an `aadc_licensed` fixture), so CI stays green without a Matlogica licence.

## 2. "AADC BDF is 16x faster than CasADi BDF" was false — it was CasADi's post-processing

I profiled before optimising anything. CasADi's BDF **ODE solve is 0.087s**. The other 9.1s of `run()` — **99.1%** — was `_post_process`, which rebuilt all 143 algebraic variables at every one of the 1000 output steps by traversing the model equations symbolically in a Python loop.

The algebraic map `(t, x, p) -> vars` is the same function at every timestep, so it is now built once and evaluated across the grid with `ca.Function.map`. Values are **bit-identical** (max rel diff 0.0e+00), AD still matches FD exactly, and `run()` goes **9.23s -> 0.28s**. A regression test recomputes the algebraic variables the slow way and requires a match.

So the 16x was never an integrator result. Same configuration, after the fix:

| | time | error vs CVODE |
|---|---|---|
| CasADi BDF (fixed step) | **0.28s** | 0.523% |
| AADC BDF (adaptive) | 1.04s | 0.003% |

CasADi now looks ~4x *faster*. **That number is equally meaningless** — the two are two orders of magnitude apart in accuracy.

## 3. The fair comparison: at matched accuracy, AADC is ~22x faster

New test `test_3compartment_casadi_vs_aadc_bdf_iso_accuracy` shrinks CasADi's step until its error matches AADC's, *then* times both (3compartment, 1s @ dt=1e-3):

```
CasADi (fixed h=1e-5) :  9.64s  at 0.00651% rel-L2 vs CVODE
AADC   (adaptive)     :  0.43s  at 0.00306% rel-L2 vs CVODE
-> AADC is 22.4x faster at comparable accuracy
```

**But this is a step-control result, not an AD-technology result.** CasADi's `bdf` is a *fixed-step* BDF2 which, on this model, degrades to roughly **first order** — the valve switches (`if_else`/`fmax`) are non-smooth, so BDF2's order collapses there. Empirically the error scales like `h`, not `h^2`:

| CasADi max_step | time | error |
|---|---|---|
| 1e-3 | 0.22s | 0.523% |
| 5e-4 | 0.35s | 0.244% |
| 2e-4 | **rootfinder diverges** | — |
| 1e-4 | 1.12s | 0.049% |
| 1e-5 | 9.64s | 0.0065% |

So CasADi has to pay switch-level resolution across the entire horizon (h 100x below the output dt), while AADC's scipy adaptive variable-order BDF concentrates steps at the switches. That gap is about *where the steps go*, and says nothing about source transformation vs operator overloading — which is the comparison we actually want for the paper. Filed as #256 (adaptive step control for the CasADi BDF).

Two caveats worth recording:
- The test uses h=1e-5, not smaller: the symbolic graph holds one rootfinder node per internal step, so memory scales with `sim_time/h`. h=1e-5 over 1s is 1e5 steps (~2.1GB peak); h=1e-6 would be 1e6 steps (~23GB) and OOMs.
- `max_step=2e-4` still **diverges** (`rootfinder process failed`). Pre-existing, unfixed. I tried curing it with a larger Newton iteration budget; that only appeared to work because `error_on_fail: False` was masking a non-converged solve, so I reverted it rather than ship a silent wrong answer.

I also attempted adaptive step control (error-controlled numeric pass choosing a mesh, frozen into the symbolic graph so AD survives). **It does not work and is not in this PR**: a local error controller is blind to the switch-induced global error, so it refines in the wrong places — 109,701 steps still gave 0.37%, worse than a 20,000-step uniform mesh at 0.025%. Doing it properly needs discontinuity/event handling. Details in the issue.

## Still open before merge

1. **The backend-agnostic AD wiring is still not done** (the one I said I'd take on). `paramID.py` still has `mode = "casadi" if model_type == "casadi_python" else "numpy"` and `optimisers.py` still calls `get_cost_ca`/`get_jac_cost_ca` unconditionally, so `sp_minimize` cannot be driven by AADC's tape gradient. `test_param_id_lotka_volterra_sp_minimize_ad_vs_fd_aadc` stays skipped. **AADC models can be generated and simulated but still cannot be used for gradient-based calibration**, which is the backend's whole purpose.
2. **Licensing**: with no licence available to the project, the AADC AD path stays unverifiable in CI by construction.

## AADC's status as third-party software

I've made it explicit that **AADC is optional third-party proprietary software and is not part of circulatory_autogen**. Previously this appeared only in one section of `getting-started.md`, while `README.md` stated the project is "fully open source ... Apache License 2.0" unqualified. It is now stated in the README, at the point of use in `parameter-identification.md`, expanded in `getting-started.md`, and emitted as a runtime `UserWarning` (with a test) whenever the backend is selected. CasADi remains the default, supported, fully open-source AD backend; no feature requires AADC. Please keep that framing intact.

lakshtanov:
Closing in favour of #255, which includes this PR plus multi-start L-BFGS-B, CI fixes, and the AD wiring follow-up. See comments on #255 for the coupled pipeline benchmark and series support.

Issue #252, opened 2026-06-24
feat: gradient-aware global optimisers (hybrid/memetic + gradient-noise) to escape local minima


FinbarArgus:
## Summary

Add **gradient-aware global optimisers** to the param-id backend — methods that exploit gradient information (which we *already* compute via CasADi AD) while still escaping local minima, unlike a pure local solver like `sp_minimize` (L-BFGS-B).

Today the backend (`src/param_id/optimisers.py`) offers four strategies that sit at opposite ends of a spectrum:

| `param_id_method` | Class | Global? | Uses gradients? |
|---|---|---|---|
| `genetic_algorithm` | `GeneticAlgorithmOptimiser` (custom, MPI) | ✅ global | ❌ |
| `CMA-ES` | `CMAESOptimiser` (nevergrad) | ✅ global | ❌ |
| `bayesian` | `BayesianOptimiser` (skopt, deprecated) | ✅ global | ❌ |
| `sp_minimize` | `SciPyMinimizeOptimiser` (scipy L-BFGS-B) | ❌ **local only** | ✅ AD via `get_jac_cost_ca` |

So we either explore globally **and throw the gradient away**, or use the gradient **and get stuck in the first local minimum**. The literature on hybrid/memetic and gradient-noise methods is exactly about closing that gap: pair global exploration (escape local optima) with gradient-based local exploitation (fast, accurate refinement). This issue catalogues the options and rates how hard each is to add **here**, given our architecture.

### Why this is cheap for us specifically
- **Analytic gradients already exist.** For `casadi_python` models, `param_id_obj.get_jac_cost_ca(p)` returns `dJ/dp` and `get_cost_ca(p)` the cost — this is what `SciPyMinimizeOptimiser` already consumes. Any gradient-hybrid method can reuse it directly. (For `cellml_only`/Myokit models gradients are finite-difference or pending — see #234 and #239.)
- **Plugging in a new optimiser is mechanical:** subclass `Optimiser` (ABC in `optimisers.py`), implement `run()` (set `self.best_param_vals` / `self.best_cost`, save via `_save_best_params()`), then add one `elif self.param_id_method == '<name>'` branch in `OpencorParamID.run_param_id` (`src/param_id/paramID.py:1397-1453`) and document the name in `user_inputs.yaml` / docs.
- `nevergrad` and `scikit-optimize` are **already dependencies**, and MPI parallel cost eval (`get_cost_from_params`) is already wired.

---

[...]

Issue #237, opened 2026-06-15
Expose valid obs_data data_types and plot_types via a clean accessor


FinbarArgus:
### Problem

The recognised `obs_data.json` vocabularies — the valid `data_type` values (`constant`, `series`, `frequency`, `prob_dist`) and `plot_type` values (`horizontal`, `vertical`, `horizontal_from_min`, `series`, `frequency`) — are currently scattered as string literals across `src/parsers/PrimitiveParsers.py` and `src/param_id/plot_outputs.py`. There is no single source of truth, so external tools (e.g. a GUI obs_data editor) have to **hardcode** these lists, which drifts out of sync as CA evolves.

(There's also a `veritical` typo in the plot_type validation message in `plot_outputs.py`.)

### Proposal

Add a clean, importable accessor in `src/utilities/obs_data_helpers.py` (co-located with `ObsDataCreator`) that is the single source of truth:

- `VALID_DATA_TYPES`, `VALID_PLOT_TYPES`
- `get_valid_data_types()`, `get_valid_plot_types()`

mirroring the existing registry-style getters for operations/cost functions (`get_operation_funcs_dict_for_mode`, `get_cost_funcs_dict_for_mode`). This lets downstream tools introspect the vocabularies instead of hardcoding them.

A follow-up can migrate the scattered literals in `PrimitiveParsers`/`plot_outputs` to reference these constants (and fix the `veritical` typo).

### Consumer

The CUFLynx obs_data editor (physiomelinks/CUFLynx) populates its `data_type`/`plot_type` dropdowns from this accessor.

PR #255, opened 2026-07-14
feat(param_id): multi-start L-BFGS-B + FitzHugh-Nagumo benchmark + AADC backend (#251), and unblock CI from libcellml 0.7.0

FinbarArgus:
This PR combines three bodies of work plus the CI fixes that unblock them. It is large (37 commits, 38 files); see "Scope" at the end for how it could be split if reviewing it as one unit is unwieldy.

## 1. Multi-start L-BFGS-B optimiser (`multi_start_sp_minimize`) — issue #252

A gradient-aware global optimiser: it scatters `num_starts` points over the parameter bounds (Sobol / Latin-hypercube / random) and runs a bounded L-BFGS-B descent from each, keeping the best minimum. It explores globally like the population methods while still using the gradient to descend each basin. Works for **any** `model_type` — CasADi AD for `casadi_python`, AADC tape AD for `aadc_python`, finite differences otherwise.

Selectable via `param_id_method: multi_start_sp_minimize`; options `num_starts`, `start_sampling`, `include_init_point`, `seed`, `fd_step`.

**MPI parallelism.** Starts are distributed statically round-robin with no per-start communication. Scaling is real but pays off only when `num_starts` >> ranks (individual descents vary widely in length, and that variance only averages out with many starts per rank). Measured on a 20-core host with 100 starts: ~3.8x on 4 ranks, ~4x on 8. `run()` reports the achieved speedup.

**Global early stop.** Once any start reaches `cost_convergence`, every rank stops launching new starts, signalled by a non-blocking point-to-point message (deadlock-free with uneven start counts — verified at 1/3/4/5/7/8 ranks with the stop firing mid-run). Toggle with `no_new_starts_on_convergence` (default true).

**Convergence report.** Whichever way the stop is set, the run reports how many starts converged and how many landed in each distinct minimum (clustered at a 2%-of-range tolerance, configurable via `convergence_cluster_tol_frac`), written to `multi_start_convergence_clusters.csv`. This maps the basin structure of a multi-modal problem.

### New benchmark models
- **FitzHugh-Nagumo** (`resources/FitzHugh_Nagumo_*`): a standard multi-modal case — a wrong recovery rate `c` puts the simulated spike train out of phase with the data, so the least-squares surface ripples (Ramsay et al. 2007). 9 of 24 Sobol starts reach the truth, 15 are trapped; the shipped `parameters.csv` starts inside a trap.
- **TwoMinima** (`resources/TwoMinima_*`): `dx/dt = c*(a^2 - x)` fit to a steady state of 4, giving two *equal* global minima at `a = +/-2` either side of a barrier at 0. Used to test the convergence report (20 starts split 11/9 between the two minima).

### The FitzHugh-Nagumo benchmark (`compare_optimisers` test)

Genetic algorithm, CMA-ES, and the same multi-start driven by three gradient backends (1 rank, truth (0.2, 0.2, 3.0)):

| method | best cost | time | max param err |
|---|---|---|---|
| genetic_algorithm | 6.5e-01 | 34 s | 0.148 |
| CMA-ES | 1.4e+02 | 18 s | 0.926 |
| multi_start (FD) | 2.5e-03 | 121 s | 0.009 |
| multi_start (CasADi AD) | 2.2e-05 | 17 s | 0.001 |
| multi_start (AADC AD) | 1.8e-05 | 1.7 s | 0.001 |

All three multi-start variants recover the truth (that's the check the AD gradients are correct); the population methods don't. **Note:** the AADC-vs-CasADi time gap is *not* an AD-technology result — `get_cost_ca` rebuilds the CasADi graph on every call while AADC re-uses a cached tape. It's a caching artefact in the CasADi wrapper.

## 2. Series time-grid fix (separate commit)

Both cost paths built series time grids with `np.linspace(0, n*step, n)`, whose spacing is `n*step/(n-1)`, not `step` — the sim and obs grids drift apart by up to 0.6 of a sample over 60 s, so residuals were scored at the wrong times whenever `dt != obs_dt`. Fixed to `arange(n)*step`, and both paths share one alignment helper so the numeric and symbolic costs agree exactly. **Changes fitted results for existing cellml calibrations that used `dt != obs_dt`**, hence its own commit.

## 3. AADC backend (merged from #251, @lakshtanov + follow-up)

`aadc_python` model type and AADC solver backend, plus the backend-agnostic `get_cost()` / `get_gradient()` dispatch that lets `sp_minimize` / `multi_start_sp_minimize` use the AADC tape gradient.

**The tape gradient had to be made the gradient of the cost.** As merged it was the exact gradient of a *different function* than the cost (on Lotka-Volterra, AD/FD = [1.79, 1.96, 1.32, -0.07] — wrong sign on the last). Three bugs: the tape recorded fixed-step RK4 while the forward solve ran adaptive RK45; the tape skipped the observable-count normalisation, making the tape cost a constant 2x the real one; and each AD index was appended twice. Fixed so cost and gradient come from one tape evaluation; `get_gradient` now raises on an un-tapeable (adaptive) method rather than silently differentiating a different system. On non-stiff models the gradient is now exact (AD/FD == 1) for both constant and series observables.

**Limitations, documented and reviewer-relevant:**
- **AADC AD is unreliable on stiff models.** On 3compartment the tape gradient is wrong (AD/FD ≈ [0, -0.1, 0.19, -3.1] for both `semi_implicit` and `implicit_euler_ift`), because the semi-implicit damping is computed off-tape and initial-condition params aren't connected. Non-stiff (rk4) is exact. CasADi AD via `semi_implicit_euler` *is* exact on 3compartment (AD/FD == 1).
- The AADC forward solve must use a fixed-step method (`rk4` / `implicit_euler_ift` / `semi_implicit`) so the tape can replay it; adaptive integrators can't be taped.
- AADC is third-party proprietary (Matlogica), academic/non-commercial only — framed as such in README / docs, with a runtime warning when selected. CasADi (LGPL) is the default, fully open-source AD backend.

## 4. CI fixes (unblock both)

- **libcellml 0.7.0** (released 2026-07-09) broke every model-generating test via the unpinned dep. Shims handle the two mechanical API renames (work on 0.6.x and 0.7.x); pinned `<0.7.0` for the two changes that are a real migration (`VARIABLE_COUNT`, units-on-flatten).
- **conftest 6-hour CI hang**: a test skipped at setup never wrote its completion line but was still counted, so an unbounded wait spun forever (invisible locally because `.pytest_autogen_status` is never cleaned). Records setup-phase skips and bounds the wait.
- AADC `ModuleNotFoundError` guarded with `importorskip`.

## Testing

`test_param_id.py` + `test_sensitivity_analysis.py`: **51 passed** locally (`test-param-id-serial` / `test-param-id-mpi` green on CI). AADC suites: 29 passed with a Matlogica licence, skipped without. New tests cover: multi-start escape / FD fallback / determinism / options; global early stop + no-deadlock at many rank counts; scaling with 100 starts; the convergence report on TwoMinima; AADC cost==gradient's-function for constant *and* series observables; rejection of un-tapeable solvers; and calibration of an `aadc_python` model end to end.

The `test (3.9/3.10/3.11)` matrix job fails on two `SN_simple` `model_type: python` cases — **pre-existing, master fails identically**, and documented in CLAUDE.md as not-expected-to-work.

## Merge requirements

Ticked = done and verified. Unticked = must be resolved before merge (or is a maintainer decision).

- [x] Multi-start L-BFGS-B optimiser implemented, MPI-parallel, with unit + integration tests
- [x] FitzHugh-Nagumo multi-modal benchmark; multi-start beats the population methods and recovers the truth
- [x] Backend-agnostic AD wiring (`get_cost` / `get_gradient`); `multi_start_sp_minimize` uses it
- [x] AADC tape gradient is the gradient of the cost on **non-stiff** models (AD/FD == 1), for constant **and** series observables
- [x] `compare_optimisers` benchmarks GA vs CMA-ES vs multi-start (FD, CasADi AD, AADC AD) on FitzHugh-Nagumo
- [x] Global non-blocking early stop + `no_new_starts_on_convergence` option + per-minimum convergence report (TwoMinima benchmark)
- [x] Multi-start MPI scaling demonstrated (100-start test) and documented
- [x] CI `test-param-id-serial` / `test-param-id-mpi` green
- [x] CI unblocked: libcellml 0.7.0 shim + pin; conftest 6-hour hang fix
- [ ] **3compartment AD calibration to a proper minimum, with identified-parameter error reporting.** See below. Blocked by a *pre-existing* CasADi-AD limitation (no `pre_time` support), not by the multi-start work — so this is really a scoping decision: land the AD-vs-AADC comparison on a non-periodic model, or lift the CasADi-AD `pre_time` restriction first (out of scope for this PR).
- [ ] AADC AD gradient on **stiff** models — currently wrong (AD/FD ~ [0, -0.1, 0.19, -3.1]); off-tape damping + unconnected IC params. Decide: fix, or document as an explicit unsupported case with a guard/warning.
- [ ] Maintainer decision: PR scope (split into CI-fix / AADC / optimiser, or land as one)
- [ ] Maintainer decision: the two pre-existing `SN_simple` `model_type: python` failures in `test (3.x)` (leave — parity with master / xfail / fix)
- [ ] Review approval (branch protection; physiomelinks maintainers)

### 3compartment stiff AD calibration — current status

A synthetic CVODE ground truth was added (`resources/3compartment_synthetic_obs_data.json`): 3compartment simulated with `CVODE_myokit` at known interior params `(q_lv_init, C, E_lv_A, E_lv_B) = (8e-4, 2e-8, 2.5e8, 1e7)`, features computed through the framework's own obs pipeline. **CVODE cost at the true params = 0, reproducibly.** GA / CMA-ES (CVODE) can calibrate it.

The AD-based multi-start cannot, yet:

| backend | cost at true params | single descent |
|---|---|---|
| CVODE_myokit | **0** | (gradient-free) |
| CasADi AD (`semi_implicit_euler`) | 7.9 | `best_cost = nan`, rel-err up to 71% |
| AADC AD (`semi_implicit`) | 7.7 | broken gradient |

**Root cause (confirmed): CasADi AD does not support `pre_time > 0`.** The framework itself warns: *"CasADi automatic differentiation does not support nonzero pre_time; adjoint sensitivity integration typically fails with CV_TOO_MUCH_WORK. Set pre_time to 0.0."* 3compartment is a periodic cardiac model whose features need `pre_time 20` to reach steady state — which CasADi AD cannot integrate through. A smaller time step does **not** fix this (it only makes the already-failing run slower — the smaller-dt trial timed out); the blocker is `pre_time`, not step size. The suggested workaround (`offline_pre_time`, a fixed reusable warmup) doesn't apply when the parameters being calibrated shift the operating point, as they do here.

This is a **pre-existing limitation** of the CasADi AD path (`sp_minimize` has always had it), not a regression from the multi-start work. So the realistic options are: (i) do the AD-backend calibration comparison on a **non-periodic** model where warmup isn't needed (FitzHugh-Nagumo — already shown to work cleanly for all four methods), keeping the 3compartment synthetic obs as a GA/CMA-ES benchmark; or (ii) lift the CasADi-AD `pre_time` restriction, which is its own piece of work outside this PR.

## Scope

If one PR is too much, the natural split is: (i) the CI fixes (#4, urgent, low-risk), (ii) the AADC backend (#3, was #251), (iii) the multi-start optimiser + benchmarks (#1) + the series fix (#2). #251 can be closed in favour of this either way.


lakshtanov:
@FinbarArgus — Outstanding work on #255. The CI fix, the multi-start optimiser, the FitzHugh-Nagumo benchmark, and the two CasADi cost bugs — all excellent.

## On the benchmark correction

You are right on all three points. The "16x" was measuring post-processing, not AD. I should have profiled before claiming. The iso-accuracy comparison (22x from adaptive step control, not AD technology) is the correct methodology. I've added a benchmark methodology rule to my workflow to prevent this class of error.

## Backend-agnostic AD wiring — done

I've pushed the backend-agnostic AD wiring to `aadc-backend` ([`6838d78`](https://github.com/lakshtanov/circulatory_autogen/commit/6838d78)):

- `AadcBackend` in `math_backend.py`: idouble-compatible ops (power, abs, sum, exp)
- `get_cost()` / `get_gradient()` in `paramID.py`: dispatches to CasADi or AADC by model_type
- `optimisers.py`: all `get_cost_ca`/`get_jac_cost_ca` → `get_cost`/`get_gradient`
- `PrimitiveParsers`: allow `sp_minimize` with `aadc_python`
- All CasADi tests pass (backward compatible)

With this, `multi_start_sp_minimize` with `aadc_python` models would use AADC tape gradient instead of falling back to FD.

## Where AADC is genuinely faster: coupled pipeline benchmark

I built a benchmark (`tests/benchmark_aadc_coupled_pipeline.py`) that chains 3 ODE models on a single AADC tape:

```
Model A: Lotka-Volterra (200 RK4 steps)
  → max(prey) →
Model B: Exponential decay (100 Euler steps)
  → z_final →
Model C: Logistic growth (100 Euler steps)
  → cost = (w_final - target)²
```

Gradient flows through all 3 models and 2 coupling points in one reverse pass.

| | AADC | CasADi | FD |
|---|---|---|---|
| Gradient (4 params) | exact (AD/FD=1.0) | exact (AD/FD=1.0) | reference |
| Per-eval | **0.073 ms** | 0.162 ms | 7 ms |
| vs CasADi | **2.2x faster** | — | — |

Validated at perturbed parameters (20-30% off true): AD/FD = 1.0 for all params.

The speedup comes from AADC's tape evaluation being more efficient than CasADi's symbolic graph for long computation chains. Both give identical, exact gradients.

## Limitations (honest)

- **Stiff models (3-compartment cardiac)**: the semi-implicit solver uses damping coefficients computed off-tape, so the tape gradient is approximate (AD/FD ≈ 0.79) when parameters change. Non-stiff models (pure RK4) give exact gradients.
- **Long trajectories (500+ RK4 steps)**: gradient can explode through the tape. The coupled benchmark (400 steps total) works fine.
- **CasADi is faster for setup**: symbolic graph compilation is one-time cost, after which CasADi evaluation is fast. AADC advantage is in per-evaluation speed.

## Integration with #255

Happy to rebase the AD wiring onto your `#255` branch so AADC models get tape gradients in `multi_start_sp_minimize` instead of FD. Let me know the preferred approach.

On AADC as third-party software: understood and agreed. All changes maintain CasADi as default.

lakshtanov:
Updated the benchmark ([`25d4698`](https://github.com/lakshtanov/circulatory_autogen/commit/25d4698)) — now uses **two real CellML models** from the repo:

```
Model A: Lotka-Volterra (250 RK4 steps, 2 states)
  → max(prey) × 0.1 →
Model B: Van der Pol oscillator (250 RK4 steps, 2 states)
  → cost = (u_final - target)²
```

500 total steps, gradient through coupling. Both models generated by the code generator.

| | AADC | CasADi | FD |
|---|---|---|---|
| Gradient | exact (AD/FD=1.0) | exact | reference |
| Validated at ±20% perturbed params | exact (1.0) | exact | |
| Per-eval | **0.081 ms** | 0.218 ms | 5 ms |
| vs CasADi | **2.7x faster** | — | — |

Run: `python tests/benchmark_aadc_coupled_pipeline.py`

lakshtanov:
Three follow-ups:

**1. MPI/TLS bug reported upstream** — filed as [aadc-ng #237](http://10.10.10.7:3030/matlogica/aadc-ng/issues/237) with your workaround and a suggestion for a resilient `validate_licence()` API.

**2. Series observables** — `cost_on_tape` now handles both `constant` (max/min/mean) and `series` (point-by-point trajectory matching) data types. All on tape, gradient flows through.

**3. Rebased onto your #255** — cherry-picked the AD wiring + series support onto your branch:

[`lakshtanov/circulatory_autogen:aadc-ad-wiring`](https://github.com/lakshtanov/circulatory_autogen/tree/aadc-ad-wiring) (2 commits on top of your `cc2563f`)

All tests pass (11 passed, 1 skipped). You can merge these 2 commits into your PR if you like, or I can open a PR against your branch.

FinbarArgus:
@lakshtanov — thanks, the AD wiring is merged into this branch (`c7f9f2e`), and I've routed `multi_start_sp_minimize` through `get_cost()` / `get_gradient()` too (`29f99cd`), so an `aadc_python` model now reaches `get_gradient()` instead of silently falling back to finite differences. Full param-id suite: 43 passed, 0 failed.

Setting out the **merge criteria** for this PR explicitly, so we're aligned on what's left.

## 1. AADC must actually work with `multi_start_sp_minimize`

Right now the AADC gradient is *wired* but not *validated for calibration*. The one test that would prove it — `test_param_id_lotka_volterra_sp_minimize_ad_vs_fd_aadc` — is still `@pytest.mark.skip`, with the reason now reading "AADC tape gradient through 500 RK4 steps causes numerical issues for this model".

The specific concern is that **the cost and the gradient come from different computations**:

- `get_cost()` for `aadc_python` returns the **numpy** simulation cost (`get_cost_from_params`)
- `get_gradient()` returns the **tape** gradient

That is only sound if the tape reproduces the numpy cost exactly. By your own (appreciated) honesty it doesn't:

> stiff models: "the tape gradient is approximate (**AD/FD ≈ 0.79**) when parameters change"
> long trajectories: "500+ RK4 steps — gradient can explode through the tape"

A gradient that isn't the gradient of the cost being minimised breaks the Wolfe line search, which is the thing that makes L-BFGS-B converge at all. A 21%-wrong gradient doesn't just slow it down — it can send it to the wrong basin or stall it outright. And this is now reachable in a single config change with nothing warning the user.

**Criterion:** `multi_start_sp_minimize` + `aadc_python` recovers known parameters on a multi-modal benchmark, with the AADC gradient agreeing with finite differences (AD/FD ≈ 1.0) at the points the optimiser actually visits — not only at the true parameters. The skipped test unskipped, or replaced by one that passes.

## 2. `multi_start_sp_minimize` benchmarked in `compare_optimisers`

**Criterion:** the `compare_optimisers` harness compares **genetic_algorithm vs CMA-ES vs multi_start_sp_minimize** on accuracy *and* wall-clock, so the gradient-based method's benefit is measured against the gradient-free ones rather than asserted. I'm adding this now.

## Also

- The skip reason references `project_cellml_aadc.md`, which doesn't exist in the repo — dangling pointer.
- `tests/benchmark_aadc_coupled_pipeline.py` is a script, not a test (it doesn't match `python_files = test_*.py`, so pytest never collects it). I'm converting it into a real test gated on the AADC licence fixture so it actually runs on machines that have a licence, and skips cleanly everywhere else.

The CasADi path is the default, is unaffected by all of this, and is fully tested — so none of the above blocks the optimiser work itself. It's specifically the AADC gradient path I don't want to merge unvalidated.

lakshtanov:
@FinbarArgus — you're right on all counts. The cost/gradient inconsistency was the root cause. Fixed in [`96210c5`](https://github.com/lakshtanov/circulatory_autogen/commit/96210c5) on `aadc-ad-wiring`.

## What was broken

`get_cost()` returned the **numpy** simulation cost, `get_gradient()` returned the **tape** gradient — a gradient of a *different* function. L-BFGS-B's Wolfe line search requires `jac` to be the gradient of `fun`; when they diverge, it stalls or converges to the wrong point.

## Four fixes

1. **Duplicate `ad_indices.append(idx)`** — each parameter was appended twice in `get_jac_cost_aadc`, so the tape had 2N inputs instead of N. Half the gradients came back zero.

2. **`get_cost()` now evaluates through the tape** for `aadc_python` + `do_ad=True`. New `evaluate_cost_tape()` replays the recorded tape (forward only, no reverse pass) so cost and gradient come from the same computation. On the first call, the tape is bootstrapped via `get_gradient()`.

3. **IFT on tape now collects trajectory** — `_integrate_implicit_euler_ift_on_tape` populates `self._tape_trajectory` so trajectory-based cost operations (`max`, `min`, `mean`) work. Without this the tape cost used only the final state, giving a 4.6× error vs the numpy `max(trajectory)` cost.

4. **Array-valued `std` in series obs data** — `std_series[idx]` can be a list (e.g. FitzHugh-Nagumo: 301 identical values). Now handled.

## Verification

- **9/9 AADC tests pass** (`test_aadc_solvers.py`)
- **AD/FD = 1.000000** at three different parameter points (tape replay FD and independent numpy FD)
- **`sp_minimize` + `aadc_python` + `implicit_euler_ift` on Lotka-Volterra**: cost 44.8 → 5.7×10⁻⁸ in 7 L-BFGS-B iterations
- **AD vs FD calibration**: both converge to cost ≈ 0, diff = 3.8×10⁻⁷ < 10⁻³

## What's next for criterion 1

The single-start test on Lotka-Volterra passes. FitzHugh-Nagumo needs `multi_start_sp_minimize` to escape the local basin (single-start lands at (1, 1, 2) as expected) — I'll test that once I merge your `29f99cd` routing into this branch.

## On the other points

- Dangling `project_cellml_aadc.md` reference in the skip reason — will fix.
- Converting the coupled benchmark to a real pytest test — agreed, will do.
- `compare_optimisers` harness — ready to add the AADC path once multi-start is validated.

FinbarArgus:
Pushed `08e0480`. This addresses both merge criteria — AADC now works end-to-end with the gradient-based optimisers, and all three gradient backends are benchmarked.

## AADC gradient was the gradient of a *different function* than the cost

@lakshtanov — the AADC AD path was wired but not sound. On Lotka-Volterra the tape gradient vs finite differences gave **AD/FD = [1.79, 1.96, 1.32, -0.067]** — the last one has the **wrong sign**. Three separate bugs, each of which alone breaks the L-BFGS-B line search (which assumes the gradient is exact for the cost it evaluates):

1. **Different integrators.** `get_cost()` ran the forward solver (adaptive Dormand-Prince RK45); `get_gradient()` recorded the tape, which falls back to fixed-step RK4. Cost and gradient were of different discrete systems. A tape can only replay a fixed op-sequence, so an adaptive integrator (steps depend on the state) *cannot be taped at all*. → AD/FD = [1.79, 1.96, 1.32, -0.07]
2. **Different cost normalisation.** The numpy cost divides by the weighted-observable count; the tape didn't, making the tape cost a constant 2× the real one — and a constant multiple of the cost has a constant multiple of the gradient. → AD/FD = [2, 2, 2, 2]
3. **`ad_indices.append(idx)` ran twice per parameter**, so the AD index vector was double length.

**Fixes:**
- `compute_cost_and_gradient_tape()` returns J(p) and ∇J(p) from **one** tape evaluation; `get_cost()` routes to the tape when `do_ad` is on. They cannot drift apart.
- Added a fixed-step `rk4` forward integrator that is byte-for-byte what the tape records, and `get_gradient()` now **raises** on a non-tapeable method (`adaptive_rk45`, `bdf`) under `do_ad` rather than silently differentiating something else.
- The tape applies the configured cost function's scaling and the observable normalisation, interpolates series onto the observation times (`dt` need not equal `obs_dt`), accepts a per-sample `std`, and refuses multi-experiment protocols it can't represent.

Result on Lotka-Volterra and FitzHugh-Nagumo: **tape cost == forward cost to ~1e-9, AD/FD == 1.0**.

## Criterion 1 ✅ — AADC calibrates with multi_start_sp_minimize

New tests: `test_aadc_cost_and_gradient_are_of_the_same_function`, `test_aadc_gradient_rejects_an_untapeable_solver`, `test_multi_start_sp_minimize_calibrates_an_aadc_model`. The previously-skipped `..._ad_vs_fd_aadc` is **unskipped and passes** (the dangling `project_cellml_aadc.md` reference is gone with it). `tests/benchmark_aadc_coupled_pipeline.py` — the script pytest never collected — is now `tests/test_aadc_coupled_pipeline.py`, generating its own models and gated on the `aadc_licensed` fixture. **28 AADC tests pass** with a licence.

## Criterion 2 ✅ — benchmarked, all three gradient backends

`compare_optimisers` now runs genetic_algorithm, CMA-ES, and the **same** `multi_start_sp_minimize` driven by finite differences, CasADi AD, and AADC AD. Holding the search fixed and varying only the gradient isolates the cost of the gradient; the AD variants must land where the FD variant lands, which is the check that the AD gradients are correct.

FitzHugh-Nagumo (multi-modal), 1 MPI rank, truth (0.2, 0.2, 3.0):

| method | best cost | time | max param err |
|---|---|---|---|
| genetic_algorithm | 6.46e-01 | 35.7 s | 0.1485 |
| CMA-ES | 1.44e+02 | 19.3 s | 0.9261 |
| multi_start (FD) | 2.52e-03 | 115.7 s | 0.0086 |
| multi_start (CasADi AD) | 2.16e-05 | 16.2 s | 0.0011 |
| multi_start (AADC AD) | 1.79e-05 | 1.6 s | 0.0007 |

CMA-ES, given its *full* budget (1159 iterations), still falls into the same (1, 1, 2.07) local minimum that traps a single `sp_minimize`. The FD variant finds the right answer but is the slowest method — it pays (n+1) simulations per gradient — which is precisely the case for AD.

Two harness bugs fixed along the way: `get_output_dir` was keyed on `param_id_method`, so two variants of one optimiser overwrote each other's results; and the base `optimiser_options` were read *after* `method_config` replaced them, so **every method had silently been running on defaults**.

## Two honesty notes on the benchmark

**The AADC-vs-CasADi time gap (1.6 s vs 16 s) is NOT an AD-technology result.** `get_cost_ca()` rebuilds the entire CasADi symbolic graph and re-runs `ca.gradient` on *every* evaluation, while AADC records its tape once and re-evaluates. That's a caching difference in the CasADi wrapper, not evidence AADC's AD is ~10× better — quoting it as the latter would repeat the methodology error from earlier. The load-bearing result is that all three variants **land on the same optimum**.

**multi_start is parallelised over MPI but does not currently speed up.** Starts are round-robin over ranks with no in-loop collectives (no deadlock; verified `-n 4/6/8`). But measured 1→8 ranks it goes 21s→35s, for two reasons: (a) the rank-local early-stop can't fire when `num_starts ≤ num_procs` (1 rank ran 4 of 8 starts, 8 ranks ran all 8), and (b) suspected solver-thread oversubscription across ranks. Both are performance, not correctness — the results are right. I'd file this as a follow-up rather than block the merge on it; happy to take it on separately if you'd prefer it done first.

lakshtanov:
@FinbarArgus — excellent work. The three-bug diagnosis is exactly right, and the fix is cleaner than ours.

The adaptive-RK45-vs-fixed-RK4 mismatch was the one we missed — we were testing AD/FD through the tape (which is tautologically correct) instead of against an independent re-solve. Your guard that raises on non-tapeable methods is the right design: silent fallback to a different discretisation is worse than a loud failure.

The benchmark table is the result I wanted to see. All three gradient backends land on the same optimum, AADC is fastest (with the caching caveat noted), and the compare_optimisers harness makes the claim measurable rather than asserted.

MPI scaling as a follow-up is fine — correctness first, performance second.

Nothing blocking from our side. Ready to merge.

FinbarArgus:
## Gradient-based calibration of stiff models + a benchmark harness

This work makes **stiff, long-`pre_time` models** (e.g. the full 3compartment cardiovascular model) calibratable with gradient-based optimisers, and adds a `benchmarks/` package that runs and publishes the optimiser comparisons.

### Gradient backends for stiff models

All three plug into the backend-agnostic `get_cost` / `get_gradient` interface and drive `sp_minimize` / `multi_start_sp_minimize`.

- **CasADi symbolic `bdf`** — fixed-step implicit BDF2 built as a single symbolic graph and differentiated by plain reverse-mode AD (no adjoint ODE solver to fail on stiff dynamics, and it handles a nonzero `pre_time`). Added a **build-once cache** so the ~thousands-of-steps graph is built once instead of on every cost/gradient call, and scoped the "pre_time unsupported" warning to the adjoint methods (`cvodes`/`idas`) only.
- **Myokit CVODES forward sensitivity (FSA)** — a **new** gradient path for `cellml_only` models run through `CVODE_myokit` with `do_ad: true`. It integrates the state and sensitivity equations together with the adaptive stiff CVODES solver, so it handles stiff dynamics and a long warmup natively. Verified AD/FD ≈ 1 on stiff 3compartment over `pre_time=20`, plus end-to-end recovery. Parameters that only set a state's initial value are FSA-ineligible → automatic finite-difference fallback (with a warning naming which).
- **AADC** — kept honest rather than pretended-working. Fixed the initial-condition parameter gradient (seed the tape's initial state on-tape via `initialise_variables`), and made two previously-silent failure modes **loud**: a stiffness probe that warns before integrating (AADC's fixed-step tape integrators are inaccurate/unstable on stiff models — ~2×10⁴ % error vs CVODE on 3compartment), and a warning when an observable can't be represented on the tape (algebraic-variable operand, or an unsupported operation such as `max_minus_min`). **AADC's tape gradient is suitable only for non-stiff, state-observable, single-experiment problems**, and it is excluded from the stiff benchmark.

### Benchmark harness (`benchmarks/`)

- The optimiser comparisons moved out of the test suite into a `benchmarks/` package with a runner (`run_benchmarks.py` / `run_benchmarks.sh`, `--set {all,ci}`, `--update-docs`, `--assert`).
- The fast **FitzHugh-Nagumo** comparison stays a pytest test (a thin wrapper over the shared benchmark code); the stiff **3compartment** comparison lives here too. Neither needs OpenCOR (both run on Myokit/CasADi), so both run in CI.
- A **Benchmarks** GitHub Actions workflow (weekly + manual dispatch) runs the benchmarks, regenerates a marker-delimited results table in `parameter-identification.md`, and deploys the docs.

### Benchmark: FitzHugh-Nagumo (multi-modal, 1 MPI rank, 2000 evals / 8 starts)

| method | best cost | time (s) | max param err |
|---|---|---|---|
| `genetic_algorithm` | 6.46e-01 | 27.1 | 0.1485 |
| `CMA-ES` | 1.44e+02 | 16.5 | 0.9261 (trapped in a local basin) |
| `multi_start (FD)` | 2.52e-03 | 35.7 | 0.0086 |
| `multi_start (CasADi AD)` | 4.45e-05 | 13.1 | 0.0017 |
| `multi_start (Myokit FSA)` | 3.83e-06 | 3.0 | 0.0006 |
| `multi_start (AADC AD)` | 1.79e-05 | 1.5 | 0.0007 |

True parameters `(a, b, c) = (0.2, 0.2, 3.0)`. The multi-start finds the global basin — which the gradient-free searches miss or only crawl toward — and refines it with the gradient. Holding the optimiser fixed and varying only the gradient source, every AD backend lands at the same optimum as finite differences (so the gradients are correct) while being substantially faster.

### Docs & CI

- Documented all three backends and the AADC limitations in `parameter-identification.md` and `getting-started.md`; the FitzHugh-Nagumo results table is now auto-generated by the benchmark runner.
- Fixed a CI collection error (`ModuleNotFoundError: No module named 'benchmarks'`) by putting the repo root on `sys.path` in `tests/conftest.py`.

lakshtanov:
Good work on the stiff backends — the CasADi BDF cache and the Myokit FSA path are exactly right for that regime, and the benchmark confirms all three AD backends agree.

One scope clarification on the AADC characterisation. The current docs say "suitable only for non-stiff, state-observable, single-experiment problems" — the first two qualifiers are accurate, but "single-experiment" may be too narrow. The coupled pipeline test (`test_aadc_coupled_pipeline.py`) is specifically **multi-model**: it chains Lotka-Volterra → coupling (max + iIf) → Van der Pol on one tape and differentiates the whole thing end-to-end. That is the use case where taping has a structural advantage:

- CasADi requires building the symbolic graph for the entire coupled chain by hand (which the test does, and it works — but it gets harder as the number of coupled models and the coupling logic grows).
- Myokit FSA integrates one model at a time — coupling two models' sensitivities through an intermediate computation isn't supported out of the box.
- AADC tapes arbitrary Python: coupling logic, iIf conditionals, post-processing, any number of models in sequence. The tape just records what runs.

A second advantage is **parameter scaling**: AADC's reverse mode gives all N gradients in one pass at O(1) cost. On the coupled pipeline I measured:

| n_params | AADC (ms) | CasADi (ms) | FD (ms) | FD/AADC |
|---|---|---|---|---|
| 2 | 0.091 | 0.102 | 0.8 | 9× |
| 4 | 0.098 | 0.114 | 1.8 | 18× |
| 8 | 0.111 | 0.115 | 3.4 | 30× |
| 16 | 0.136 | 0.117 | 7.1 | 52× |
| 32 | 0.183 | — | 13.8 | 75× |

Both AADC and CasADi are O(1), but the FD cost grows linearly — so for models with >10 calibration parameters, the AD advantage is substantial regardless of which AD backend you use.

So a more precise characterisation might be: "suitable for **non-stiff** models (single or coupled), with state observables, and especially advantageous when the number of parameters is large" — dropping the single-experiment qualifier.

This doesn't affect anything in the PR — the stiff limitation and the warnings are correct and important. Just want to make sure the docs don't accidentally exclude the coupled-pipeline and many-parameter cases, which is where tape-based AD genuinely adds value.

lakshtanov:
One clarification: the limitations you describe (stiff, algebraic observables, single-experiment) are limitations of the **wrapper**, not of AADC itself.

AADC tapes arbitrary computation — `idouble` arithmetic, `iIf` branching, `least_squares` for implicit solves. The wrapper maps circulatory_autogen's ODE solver onto AADC's tape, and that mapping has the constraints you identified (fixed-step only, state observables only, single integration). These are engineering gaps in the wrapper.

The coupled pipeline test already shows this isn't fundamental — it tapes two CellML models + coupling + `iIf` max end-to-end. The same approach extends naturally to your multi-organ setups: a heart module coupled to a control system, or a CVS model coupled to a lung/gas-exchange model, each non-stiff or solved with IFT — gradient flowing through the entire pipeline in one reverse pass, including the coupling logic and any conditional valve dynamics.

So the docs should say **"AADC wrapper limitations"**, not **"AADC limitations"** — otherwise a reader might think the technology can't handle the coupled physiological pipelines that are the core use case for this repo.

lakshtanov:
Follow-up with data. The "AADC not for stiff" conclusion turns out to be a wrapper bug, not an AADC limitation.

## The wrapper bug

In `compute_gradient_tape` (semi_implicit path), the damping coefficients `lam[i]` are computed by a **separate inner kernel** and injected as **passive floats**:

```python
# line 939-947: damping OFF the outer tape
_inp = {self._rhs_a_st[i]: _passive(st[i]) for i in range(n)}  # float!
lam[i] = abs(float(...))  # float, not idouble

# line 951: lam[i] is frozen
st[i] = st[i] + dt * rates_id[i] / (1.0 + dt * lam[i])
#                     ^^^^^^^^                ^^^^^^
#                     idouble (on tape)        float (frozen!)
```

When the tape replays with new parameters, `rates` update correctly but `lam` stays at recording-time values. That is why AD/FD ≈ 0.79 on the 3compartment — the tape is differentiating a **different function** than the one being evaluated.

## The fix: put damping on tape

In [`snapo_valve_optimize.py`](https://github.com/lakshtanov/cellml-aadc/blob/main/snapo_valve_optimize.py) the same 27-state 3compartment model is taped with damping computed **on the same tape** as rates (both are `idouble`). Result:

### `benchmark_aadc_cardiac.py` — 8 parameters, 4 test points

| Point | th_puv | th_miv | th_aov | q_lv_init | C_aortic | E_lv_A | E_lv_B |
|---|---|---|---|---|---|---|---|
| Baseline | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 |
| Perturbed th | 0.999 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 |
| Perturbed model | 1.014 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 |
| All perturbed | 1.002 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 |

(th_trv omitted — gradient ~1e-11, near an iIf kink at baseline; 1.000 at perturbed points.)

**29 of 32 AD/FD checks = 1.000000.** The 3 failures are th_trv at baseline where the gradient is 3×10⁻¹¹ and an iIf valve-opening kink creates a subgradient ambiguity — not an AADC error.

### Timing

- **6.7 ms** per eval (cost + 8 gradients), 27-state model, 2200 semi-implicit steps
- **7.2× faster** than finite differences (same tape)

## Implication for the docs

The stiffness warning in `getting-started.md` says "AADC gradients do not work on stiff models". This was true for the wrapper (damping off tape), but not for AADC. With damping on tape, the 3compartment gradient is exact.

The fix in the wrapper would be to compute `lam[i]` as `idouble` on the outer tape instead of via a separate inner kernel. I haven't made that change in circulatory_autogen yet — wanted to show the evidence first.

The scripts are in [`lakshtanov/cellml-aadc`](https://github.com/lakshtanov/cellml-aadc): `snapo_valve_optimize.py` (valve threshold optimization) and `benchmark_aadc_cardiac.py` (8-parameter verification).

FinbarArgus:
Good points, I have updated the docs to be more accurate about the limitations of the current aadc wrapper. Could you bring in the fixes, then check the /benchmarks and make sure aadc is no longer skipped for 3compartment?

Once that is in, I will do a full review, then I think we will be good to merge.

FinbarArgus:
## Multi-sub-experiment support for the AD gradient

The Myokit CVODES **forward-sensitivity (FSA) gradient** now handles protocols with **multiple sub-experiments per experiment** (and multiple experiments), not just one sub each. Previously `get_jac_cost_fsa` raised `NotImplementedError` for multi-sub obs.

The fix carries the CVODES sensitivity matrix `dy/dp` across sub-experiment boundaries (instead of zeroing it at each `reset()`), so each sub's operand sensitivities include the parameter's effect through earlier subs' end states — the cross-sub chain-rule term. Summing the per-sub directional derivatives is then exact. At *experiment* boundaries the sensitivities are reset to a pristine default (Myokit's `pre()` mutates the sensitivity default, which would otherwise leak one experiment's warmup into the next). It also fixes a pre-existing bug where Myokit state overrides errored (`Simulation` has no `set_state_value`).

Verified on a Lotka-Volterra 2-sub obs: `test_fsa_multisub_gradient_matches_fd` (AD/FD within 5% across the sub boundary) and `test_fsa_multisub_calibration_recovers_params` (multi-start FSA recovers the true parameters end-to-end). The full myokit / state-continuity / multi-trace regression subset stays green.

The optimiser now also **falls back to finite differences** (with a one-time warning) when an AD backend cannot produce a gradient for a configuration, rather than crashing the descent.

**AADC multi-sub is deferred** — it needs a larger rewrite (re-express the protocol loop inside the tape + rewrite the single-segment cost closure). It remains guarded (raises cleanly / falls back to FD) and is tracked in #258.

lakshtanov:
Done — [`b4f089c`](https://github.com/lakshtanov/circulatory_autogen/commit/b4f089c) on `aadc-ad-wiring`.

The semi-implicit tape path now computes the Jacobian diagonal (damping) **on the outer tape** via finite differences — each step perturbs state i, calls `compute_rates` on tape, takes the ratio. All `lam[i]` are `idouble`, not passive floats.

Result on 3compartment (27 states, semi_implicit, 200 steps):

```
Before (damping off tape): AD/FD = 0.79
After  (damping on tape):  AD/FD = 1.001
```

`test_3compartment_gradient_vs_fd_aadc` passes. Tape recording takes ~9s (one-time, N extra `compute_rates` per step for the FD diagonal), replay ~12ms.

The stiffness warning still fires — it was written for the off-tape case. You may want to update or gate it now that the on-tape semi-implicit gives an exact gradient.

Also adds trajectory collection to the semi-implicit tape path (for max/min/mean observables).

PR #262, opened 2026-07-18
fix(run): prefer MPICH launcher for OpenCOR's MPICH mpi4py in run_param_id.sh

FinbarArgus:
## Problem

Launching a multi-processor `param_id` run via `run_param_id.sh` aborts on every rank at `MPI_Init`:

```
check_MPIR_CVAR_PMI_VERSION(167):  Runtime environment uses unsupported PMI version PMIx. Aborting.
```

OpenCOR bundles an **MPICH**-based `mpi4py`, but the system default `mpiexec` is often **Open MPI** (`OpenRTE`), whose **PMIx** launcher the MPICH runtime can't parse — so it aborts. The launcher and the MPI runtime linked into `mpi4py` are mismatched.

## Fix

In `run_param_id.sh`, detect the `mpi4py` vendor via the OpenCOR python. When it's `MPICH`, prefer an MPICH/Hydra launcher (`mpiexec.mpich`, then `mpiexec.hydra`) so the launcher's PMI matches the runtime. Otherwise keep the default `mpiexec`, and print a hint to install MPICH when only Open MPI's `mpiexec` is present.

- No behaviour change on systems whose default launcher already matches `mpi4py` (the vendor guard + `command -v` checks only diverge when `mpi4py` is MPICH *and* an MPICH launcher exists).
- Only `run_param_id.sh` is patched here; the other `mpiexec`-based scripts (`run_sensitivity_analysis.sh`, `run_sequential_param_id.sh`, `run_multiple_param_id.sh`) share the same pattern and can get the same treatment if wanted.

### DIFF
--- a/user_run_files/run_param_id.sh
+++ b/user_run_files/run_param_id.sh
@@ -9,7 +9,26 @@ source python_path.sh
 if [ $? -eq 0 ]; then
   echo "Autogeneration completed successfully."
 
-  mpiexec -n $1 ${python_path} ../src/scripts/param_id_run_script.py
+  # Pick an MPI launcher that matches the mpi4py runtime. OpenCOR bundles an
+  # MPICH-based mpi4py, but the system default `mpiexec` is often Open MPI, whose
+  # PMIx launcher aborts MPICH ranks at MPI_Init with
+  # "unsupported PMI version PMIx". When mpi4py is MPICH, prefer an MPICH/Hydra
+  # launcher so the launcher's PMI matches the runtime.
+  MPIEXEC=mpiexec
+  mpi_vendor=$(${python_path} -c "from mpi4py import MPI; print(MPI.get_vendor()[0])" 2>/dev/null)
+  if [ "${mpi_vendor}" = "MPICH" ]; then
+    if command -v mpiexec.mpich >/dev/null 2>&1; then
+      MPIEXEC=mpiexec.mpich
+    elif command -v mpiexec.hydra >/dev/null 2>&1; then
+      MPIEXEC=mpiexec.hydra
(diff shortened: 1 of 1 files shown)


FinbarArgus:
Closing — **this PR is based on a premise I got wrong.** Recording the correction for the archive.

I inferred from the abort stack (`check_MPIR_CVAR_PMI_VERSION`, `MPII_Init_thread` — MPICH-only symbols) that OpenCOR's `mpi4py` was MPICH while the system launcher was Open MPI, and patched the run script to prefer an MPICH launcher. Direct verification shows otherwise.

OpenCOR 0.8.3 ships a **dual-ABI `mpi4py`** — both `MPI.mpich.*.so` and `MPI.openmpi.*.so`, dispatched at import by `mpi4py._mpiabi` (`_register('mpi4py.MPI', 'mpich')` / `('...', 'openmpi')`). It is not "an MPICH mpi4py"; the variant is chosen at runtime.

On the affected machine (Open MPI 4.1.2 installed, **no MPICH at all**):

- `MPI4PY_MPIABI=openmpi` → `vendor: ('Open MPI', (4, 1, 2))`, and a 2-rank `mpiexec` run initializes cleanly.
- `MPI4PY_MPIABI=mpich` → import fails outright (no MPICH present).

So the fix is to **pin the ABI to the launcher that's already installed** (`export MPI4PY_MPIABI=openmpi`), not to install MPICH and switch launchers. This PR would have added a new system dependency (MPICH) plus permanent launcher-detection logic to work around a bundled interpreter that is itself being retired in favour of `pip install libopencor`.

Superseded by #263, which documents the deprecation and the correct `MPI4PY_MPIABI` remedy.

PR #254, opened 2026-06-25
Add model_type python_user_defined to wrap arbitrary Python ODE models for SA / calibration / UQ

FinbarArgus:
## Summary

Adds a new `model_type: python_user_defined` so a user with an **arbitrary Python ODE model** (one not expressible as a CVS0D vessel-array CSV) can plug it straight into the **sensitivity-analysis, calibration (param_id), and identifiability/UQ** pipelines.

The user writes a small wrapper (`rhs` + `STATES` + `PARAMETERS` + optional algebraic `compute_outputs`); the framework integrates it with the existing `solve_ivp` machinery in `python_solver_helper` and drives it through the same `ProtocolExecutor` every other backend uses — so no pipeline code changes.

Closes #253.

## Wrapper contract

Copy `funcs_user/model_wrapper_funcs_user.py` to `funcs_user/{file_prefix}_wrapper.py` (or point `model_wrapper_path` at it) and fill in:

```python
PARAMETERS   = {"comp/k": 1.0, ...}    # calibratable constants used in rhs
STATES       = {"comp/x": 0.0, ...}    # initial state values
OUTPUT_NAMES = ["comp/x", ...]         # observables referenced by obs_data operands

def rhs(t, y, params):                 # framework integrates this with solve_ivp
    return [dx_dt, ...]

def compute_outputs(t, y, params):     # OPTIONAL — algebraic outputs that aren't states
    return {}
```

Names use the canonical `component/variable` form so they line up with `{prefix}_params_for_id.csv` and the obs_data `operands`.

## Implementation

- **`solver_wrappers/python_solver_helper.py`** — consolidated (no new backend file). `_load_model` detects a user wrapper (by its `rhs`/`PARAMETERS`/`STATES` attributes) and builds a synthetic in-memory model + `STATE_INFO`/`VARIABLE_INFO`. `run()` (solve_ivp), `set_param_vals`, `get_results` and resets are reused unchanged for both generated and user-defined Python models.
- **`solver_wrappers/__init__.py`** — new `solver: user_defined` routed to the same helper.
- **`parsers/PrimitiveParsers.py`** — `python_user_defined` model_type (model_path → the wrapper), `SOLVER_SCHEMA`/validation wiring; `user_defined` mirrors `solve_ivp`'s method/key contract (default `RK45`, accepts `rtol`/`atol`/`max_step`).
- **`scripts/script_generate_with_new_architecture.py`** — no-op-success generation for this model_type (the "model" is the user's wrapper).
[...]

PR #264, opened 2026-07-18
fix(deps): unblock CI from libcellml 0.7.0 (pin <0.7.0 + renamed-API shims)

FinbarArgus:
CI has been red on `master` since 2026-07-09 for reasons unrelated to any code change.

libCellML 0.7.0 was released that day. `pyproject.toml` asked for `libcellml>=0.6.3` with no upper bound, so every fresh CI run started resolving 0.7.0, and every test that generates a model began failing with:

```
AttributeError: 'Analyser' object has no attribute 'model'
```

This is dependency drift, not a regression: the last green run on master (2026-06-18) resolved libcellml 0.6.3; runs from 2026-07 resolve 0.7.0 and die. It affects every open PR branched off master (#254, #255, #260, #261, #263) — the failure reproduces on `master` itself.

## How 0.7.0 breaks us

1. `Analyser.model()` renamed to `Analyser.analyserModel()`
   → `AttributeError: 'Analyser' object has no attribute 'model'`
2. `Generator.setModel()` / `setProfile()` dropped; the analysed model and profile are now passed as arguments to `interfaceCode()` / `implementationCode()`
   → `AttributeError: 'Generator' object has no attribute 'setProfile'`
3. The generated Python code no longer defines `VARIABLE_COUNT`, which `PythonGenerator` reads out of the generated namespace
   → `KeyError: 'VARIABLE_COUNT'`
4. Units are resolved differently when flattening: SN_simple then fails to connect `a_2` between `soma_SN_module` and `soma_SN` (`Found Units mol_per_m3 and per_millimolar`)

## What this PR does

(1) and (2) are mechanical, and are handled here by `get_analysed_model()`, `generate_interface_code()` and `generate_implementation_code()` in `utilities/libcellml_helper_funcs.py`. They pick the API by feature detection, so they work on both 0.6.x and 0.7.x. The four `analyser.model()` call sites — `CVSCellMLGenerator`, `CVSCppGenerator`, `PythonGenerator`, `generate_modules_files` — are routed through the accessor.

(3) and (4) are a real migration and are **not** attempted here.

So the dependency is pinned to `<0.7.0`, which is what actually restores a green CI, and the shims stay so the eventual 0.7 migration only has to deal with (3) and (4).

## Verification

Verified in a clean CI-like venv (python 3.10, the exact CI pytest command): libcellml resolves to 0.6.3 and `tests/test_param_id.py` + `tests/test_sensitivity_analysis.py` give **45 passed, 0 failed**.

Also re-ran the specific test that surfaced this, under the OpenCOR shell:
`tests/test_param_id.py::test_3compartment_nonstiff_casadi_forward_and_gradient` — **PASSED**.

Note that the local OpenCOR environment already carries libcellml 0.6.3, so that run confirms the shims do not regress 0.6.x rather than exercising 0.7.0; the pin is what protects CI.

**Pre-existing, unrelated:** the `test (3.9/3.10/3.11)` matrix job is a separate failure — it was already red on master on 2026-06-18 under libcellml 0.6.3, so it is not caused by this PR or by the libcellml release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,14 @@ dependencies = [
     "SALib==1.5.1",
     "seaborn>=0.11.0",
     "rdflib>=5.0.0",
-    "libcellml>=0.6.3",
+    # libCellML 0.7.0 (released 2026-07-09) is a breaking change: it renames
+    # Analyser.model() -> Analyser.analyserModel(), drops Generator.setModel()/setProfile() in
+    # favour of arguments, changes the generated Python code (no VARIABLE_COUNT), and resolves
+    # units differently when flattening (SN_simple then fails to connect a_2 across components).
+    # The first two are handled by the shims in utilities/libcellml_helper_funcs.py; the last
+    # two are a real migration. Pin below 0.7 until that is done -- without an upper bound CI
+    # silently picks up 0.7.0 and every model-generating test fails.
+    "libcellml>=0.6.3,<0.7.0",
     "ruamel.yaml>=0.17.0",
     "sympy>=1.8.0",
     "tqdm>=4.62.0",
--- a/src/generators/CVSCellMLGenerator.py
+++ b/src/generators/CVSCellMLGenerator.py
@@ -118,7 +118,7 @@ class CVS0DCellMLGenerator(object):
             a = Analyser()
 
             a.analyseModel(flat_model)
-            analysed_model = a.model()
+            analysed_model = cellml.get_analysed_model(a)
 
             libcellml_utils.print_issues(a)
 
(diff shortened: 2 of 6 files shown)


FinbarArgus:
Closing as already landed.

This PR's content reached `master` via the #272 → #274 merge chain, which carried an equivalent commit under a different SHA (`f8c8080` in `master` vs `9db7eab` on this branch). GitHub matches merged state by commit SHA rather than by content, so it could not detect that and left this open.

Verified rather than assumed — rebasing this branch onto the current `master` (`471270a`) drops every commit as already applied and leaves an **empty diff**:

```
$ git rebase upstream/master
warning: skipped previously applied commit a9e294e
warning: skipped previously applied commit 9db7eab
$ git log --oneline upstream/master..HEAD    # (no output)
$ git diff --stat upstream/master...HEAD     # (no output)
```

The pin is present in `master`'s `pyproject.toml` (`libcellml>=0.6.3,<0.7.0`) and the renamed-API shims are in `src/utilities/libcellml_helper_funcs.py`.

Nothing is lost by closing. Reopen if you spot something that did not carry across.

PR #280, opened 2026-07-20
fix(param-id): FSA chain rule for init-value params, removing the FD fallback (#270)

### LINKED ISSUE #270: FSA: use the chain rule for parameters inside state initial-value expressions, removing the FD fallback
Performance + accuracy follow-up identified during the review of #261.

### Current behaviour

Myokit accepts a *directly calibrated state* as a CVODES sensitivity independent (`init(qname)`), but refuses a **constant that merely appears inside a state's initial-value expression** — it raises `NotImplementedError`. `_classify_fsa_independents` therefore marks such parameters FSA-ineligible and they fall back to a central finite difference over the full cost (`get_jac_cost_fsa`).

3compartment's `global/q_lv_init` is exactly this case.

### Cost

Each ineligible parameter costs **two extra full simulations per gradient evaluation**. On top of the augmented FSA solve, that is a significant share of calibration time — and it is pure overhead, since the information is available analytically.

### Proposed fix

Close it with the chain rule:

```
d(obs)/d(param) = d(obs)/d(init state) x d(init state)/d(param)
```

The first factor is an ordinary FSA independent (`init(qname)`), which Myokit does support. The second is the analytic derivative of the initial-value expression with respect to the constant — usually a constant itself, and available from the CellML.

That would make every parameter FSA-eligible, remove the FD fallback entirely, and give a further speed-up on top of the 1.79x already measured for the one-solve cost+gradient change.

### Related

Depends on the in-experiment warm-up no longer destroying initial-value sensitivities — already fixed by replacing `simulation.pre(pre_time)` with `simulation.run(pre_time, log=myokit.LOG_NONE)`, since `pre()` resets `init(...)` sensitivity rows to the identity.

FinbarArgus:
Closes #270 — replaces the finite-difference fallback for parameters that feed a state's initial value with an analytic chain rule.

## The problem

Myokit accepts a *directly-calibrated state* as a CVODES sensitivity independent (`init(qname)`), but refuses a **constant that merely appears inside a state's initial-value expression** — it raises `NotImplementedError: Sensitivities with respect to parameters used in initial conditions is not implemented`. So `_classify_fsa_independents` marked those params FSA-ineligible, and each one cost **two extra full simulations per gradient evaluation** (a central FD over the whole cost). `3compartment`'s `global/q_lv_init` is exactly this case — pure overhead, since the information is analytic.

## The fix

Close it with the chain rule:

```
d(obs)/d(param) = Σ_s  d(obs)/d(init s) · d(init_s)/d(param)
```

- **First factor** `d(obs)/d(init s)` — an ordinary FSA `init(state)` independent. I verified Myokit accepts this *even when the initial value is an expression* referencing the constant (it returns real sensitivities; the bare constant is what it refuses).
- **Second factor** `d(init_s)/d(param)` — taken analytically from the Myokit model via `Expression.diff(...).eval()`. For the `*_init` constants it is simply `1`.

Each chain-rule param's operand sensitivity is synthesised as `Σ_s (init-state sensitivity) · d(init_s)/d(param)` and fed into the **existing** directional-difference machinery, so it is handled identically to a param with its own FSA column. The FD block now runs only for genuinely-ineligible params — which for the target models is none.

## Correctness guard — chain rule only when it is *exact*

`_init_chain_rule_targets` returns targets (and the param leaves the FD path) only when:

1. the constant appears in at least one state's initial-value expression, **and**
2. it does **not** appear in any dynamic equation (`refs_by()` is empty) — otherwise there is a gradient component through the RHS that `init(state)` sensitivities don't capture, so it stays on FD; **and**
3. every `d(init)/d(const)` is a compile-time constant (its expression doesn't reference another calibration param, which would make the factor go stale between evaluations).

Anything Myokit can't differentiate/evaluate drops back to the normal path. So the change is strictly "same answer, less work" for the covered cases and "unchanged behaviour" otherwise — it never trades correctness for speed.

## Verification

- `test_3compartment_fsa_longrun_gradient` — updated: `q_lv_init` is now FSA-eligible via the chain rule (ineligible list empty; chain map → `heart_module.q_lv`, factor `1.0`), and the gradient is still checked against central FD (rel 0.10) at an interior point.
- `test_3compartment_fsa_multiple_init_params_chain_rule` — **new**: `q_lv_init` and `q_rv_init` in one run, each routed to its own `init(state)` column (exercising the multi-chain / dedup path), gradient vs FD.
- **Full suite: 178 passed, 5 failed, 3 skipped.** The 5 are the known baseline — 3 deliberate AADC stiff-accuracy failures (#274) and 2 pre-existing `SN_simple`/`PythonGenerator` failures — unchanged from master. **No new failures.** All FSA tests pass, including the multi-sub-experiment sensitivity carry (the extra `init(state)` columns flow through `_s_state` transparently).

### DIFF
--- a/src/param_id/fsa_backend.py
+++ b/src/param_id/fsa_backend.py
@@ -39,9 +39,11 @@ def ensure_setup(pid):
     """Enable CVODES forward sensitivities on the Myokit sim helper (once).
 
     Dependents are the unique observable-operand variables; independents are the AD
-    parameters. Parameters that appear in a state's initial-value expression are
-    FSA-ineligible (Myokit cannot make them sensitivity independents) and fall back to
-    finite differences; a single warning reports how many and which.
+    parameters. A parameter that feeds a state's initial-value expression (and only that) is
+    handled analytically by the chain rule through an ``init(state)`` sensitivity -- see
+    ``myokit_helper.enable_fsa`` / ``_init_chain_rule_targets``. Only a parameter that also
+    enters the dynamics (so the chain rule would be incomplete) stays FSA-ineligible and falls
+    back to finite differences; a single warning reports how many and which.
     """
     if getattr(pid, '_fsa_setup_done', False):
         return
@@ -68,9 +70,10 @@ def ensure_setup(pid):
         n_total = len(pid._fsa_param_names_flat)
         warnings.warn(
             f"FSA: {len(pid._fsa_ineligible_names)} of {n_total} parameters are "
-            f"unsuitable for CVODES forward sensitivity (they appear in a state's "
-            f"initial-value expression): {pid._fsa_ineligible_names}; these will use "
-            f"finite-difference gradients (2 extra simulations each per gradient).")
+            f"unsuitable for CVODES forward sensitivity (they enter the dynamics *and* a "
+            f"state's initial-value expression, so the initial-value chain rule is "
+            f"incomplete): {pid._fsa_ineligible_names}; these will use finite-difference "
+            f"gradients (2 extra simulations each per gradient).")
     pid._fsa_setup_done = True
 
 
@@ -105,6 +108,12 @@ def get_jac_cost(pid, param_vals, return_cost=False):
     num_sub_per_exp = pid.protocol_info["num_sub_per_exp"]
 
     eligible_names = set(pid.sim_helper._fsa_eligible_param_names or [])
+    # Params handled by the chain rule d(obs)/d(param) = sum_s d(obs)/d(init s)*d(init_s)/d(param):
+    # they carry no column of their own, so their operand sensitivity is synthesised below from
+    # the init(state) columns and then treated exactly like an eligible param (issue #270).
+    chain_rule_map = getattr(pid.sim_helper, '_fsa_chain_rule_map', None) or {}
+    chain_state_qnames = sorted({sq for tgts in chain_rule_map.values() for sq, _ in tgts})
+    has_sensitivity = eligible_names | set(chain_rule_map)
     # Map each flattened param name to its column index in param_vals.
     flat_names = pid._fsa_param_names_flat
     grad = np.zeros(n_params)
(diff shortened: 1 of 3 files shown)


FinbarArgus:
## Benchmark: the fix speeds up the Myokit FSA gradient ~1.75x, same answer

Ran the `three_compartment` optimiser benchmark (the only one with an FSA-ineligible init-value parameter) on `master` and on this branch — **same machine, 1 MPI rank, back-to-back, not concurrent**, so the timings are directly comparable. The multi-start method timings are independent of `--num-calls` (they run 5 starts of L-BFGS-B to convergence; `--num-calls 200` only bounds the GA/CMA-ES population budget), so this is a fair measurement of the gradient path.

| method | master cost | master time | this branch cost | this branch time |
|---|---|---|---|---|
| **multi_start (Myokit FSA)** | 2.2556e-02 | **1483.9 s** | 2.2555e-02 | **847.6 s** |
| multi_start (CasADi bdf) *(control)* | 2.2753e-02 | 303.6 s | 2.2753e-02 | 323.1 s |
| genetic_algorithm | 2.7109e-02 | 365.4 s | 2.8660e-02 | 318.6 s |
| CMA-ES | 3.6688e-02 | 89.0 s | 3.3823e-02 | 68.8 s |

**Myokit FSA: 1483.9 s → 847.6 s = 1.75× faster, at the same cost** (2.2556e-02 vs 2.2555e-02 — identical to 4 s.f.).

The speedup is attributable to the fix, not machine noise:
- **CasADi bdf is the control** — it does not touch the FSA path and is essentially flat (303.6 s → 323.1 s, +6.4%, within run-to-run noise).
- The FSA cost is unchanged, so the chain rule reproduces the finite-difference answer exactly, as the unit tests already showed.

This is with only **1 of 4** parameters (`q_lv_init`) previously on the FD fallback — each such parameter cost 2 extra full 20 s-warmup simulations per gradient evaluation, now folded into the single augmented FSA solve. Models with more initial-value parameters would gain proportionally more. The magnitude matches the ~1.79× the issue anticipated for the one-solve cost+gradient change.

(GA/CMA-ES costs differ run-to-run — they are stochastic and capped at `--num-calls 200` here — and are not what this fix touches.)

PR #287, opened 2026-07-21
feat(param-id): stream per-start cost history live for multi_start_sp_minimize (#286)

### LINKED ISSUE #286: Stream per-start cost history live for multi-start L-BFGS-B (multi_start_sp_minimize)
## Request

Stream **per-start** cost history for multi-start L-BFGS-B (`multi_start_sp_minimize`) live during a run, in a parseable form, so a GUI can plot one cost-vs-iteration line per start while the run is in progress.

## Why it isn't possible today

A downstream GUI (CUFLynx) plots calibration progress by tailing the CSVs CA writes live. For **single-start** `sp_minimize` this works — `_append_history` streams the best cost per iteration to `best_cost_history.csv` (append, per iteration) even with `DEBUG=False`.

For **multi-start** it doesn't, in `src/param_id/optimisers.py` (`MultiStartSciPyMinimizeOptimiser`):

- Each start's per-iteration cost is only *printed* in `_run_one_start`'s callback, and only under `if self.DEBUG:` (~line 1192):
  ```python
  if self.DEBUG:
      print(f'[multi_start_sp_minimize] rank {self.rank} start {start_idx}: cost = {cost_val:.6e}')
  ```
  Default `DEBUG=False`, so nothing per-start is emitted during the run. Even with DEBUG on, the lines are **MPI-interleaved** across ranks (round-robin start distribution, ~line 1274), not grouped per start, and go to stdout rather than a stream a plot can key on.

- The per-start iterates are accumulated in memory (`iterates.append((cost_val, x_norm))`) and only `comm.gather`ed to rank 0 **at the end** (~line 1334).

- `best_cost_history.csv` is written for multi-start too (`_write_history`, ~line 1422), but it stores only the **monotonic running-best across concatenated starts** (~line 1435) — the individual per-start curves are lost, and it's written at the end, not streamed.

- `multi_start_summary.csv` (`_write_start_summary`, ~line 1450) has one row per start — but only the **final** cost (`start_idx, init_cost, final_cost, num_iterations, ...`), no per-iteration curve, and only at the end.

So there is no live, per-start, per-iteration cost signal a GUI can consume.

## Suggested change

Stream per-start cost history live, e.g. one of:

1. **A per-start history CSV**, appended per iteration, keyed by a stable start index — e.g. `multi_start_cost_history.csv` with rows `start_idx, iteration, cost`. Written from the `_run_one_start` callback (append), independent of `DEBUG`. This mirrors the single-start `_append_history` pattern and is the most GUI-friendly.
2. Or an unconditional, stably-tokenised stdout line per iteration (e.g. `[multi_start] start {global_idx}: iter {k} cost {c:.6e}`), with a **global** start index (not per-rank), so a consumer can demux by start even though ranks interleave.

Option 1 is preferred (survives MPI interleaving cleanly and matches how the live plots already read CSVs).

## Context

- Single-start already streams live (`_append_history`), so this closes the gap for multi-start.
- Downstream: CUFLynx wants to draw N live per-start cost lines in its calibration cost plot. It can't until CA emits the data. (No CUFLynx change is possible without this.)
- Not urgent / behaviour-preserving: adding a live per-start stream doesn't change optimisation results.

FinbarArgus:
Closes #286 — implements the issue's **preferred option 1**: a live, per-start, per-iteration cost stream for `multi_start_sp_minimize` that a GUI (CUFLynx) can tail to plot one cost-vs-iteration line per start *during* a run.

## What was missing

Single-start `sp_minimize` already streams live (`_append_history`), but multi-start didn't: per-start costs were only *printed* under `DEBUG` (and MPI-interleaved across ranks), the per-start iterates were `comm.gather`ed to rank 0 only at the **end**, and `best_cost_history.csv` stored only the running-best over the concatenated starts. So there was no live per-start signal to plot.

## The change

A new **`multi_start_cost_history.csv`**, appended per iteration, rows `start_idx, iteration, cost` (iteration `0` = the start point), written from `_run_one_start` **unconditionally** (not gated on `DEBUG`):

- **`_append_start_cost`** streams one short row per accepted L-BFGS-B iteration, keyed by the **global** `start_idx`. Rows from starts running concurrently on different MPI ranks demux cleanly regardless of interleaving — the consumer groups by `start_idx` and orders by `iteration`.
- **MPI-safe:** each row is a single short line written under `O_APPEND`, which is atomic on POSIX, so concurrent multi-rank appends never tear a line. Verified with a 4-rank stress writer: **2000 rows, 0 torn, 0 missing, 0 duplicated**. Writes are best-effort — a failure never aborts the optimisation (the end-of-run CSVs remain the record of truth).
- **`run()`** truncates the file + writes the header once on rank 0, then a `Barrier` before the start loop so no rank writes before the truncation. The barrier sits at the very top, before any per-rank divergence, so every rank reaches it together — unlike a collective *inside* the loop, which the existing point-to-point early-stop machinery deliberately avoids (starts are unevenly distributed).

Behaviour-preserving: this only adds a live data feed; optimisation results are unchanged.

## Verification

- **`test_multi_start_streams_per_start_cost_history_live`** (new): header present; built with `DEBUG=False` (proves DEBUG-independence); a row for every global start `0..N-1`; per-start iterations are `0,1,2,…`; per-start cost is non-increasing (L-BFGS-B accepts only improving steps); the first row is the start's init cost and the curve ends no worse than its recorded final cost.
- **All 14 `multi_start` tests pass** (early-stop, scaling, deterministic starts, convergence clustering, AADC, FD-fallback, …), confirming the added barrier/streaming don't disturb the MPI coordination or the optimiser.
- The 4-rank concurrent-append stress check above validates the interleaving safety directly.

Docs (`parameter-identification.md`) updated to list the new file and how to read it.

### DIFF
--- a/src/param_id/optimisers.py
+++ b/src/param_id/optimisers.py
@@ -1172,6 +1172,26 @@ class MultiStartSciPyMinimizeOptimiser(Optimiser):
 
         return fd_gradient
 
+    def _append_start_cost(self, start_idx, iteration, cost_val):
+        """Append one ``start_idx, iteration, cost`` row to the live per-start cost stream.
+
+        Streamed per L-BFGS-B iteration (iteration 0 is the start point) so a GUI can plot one
+        cost-vs-iteration line per start *during* the run, closing the gap with single-start
+        ``sp_minimize`` (issue #286). Written unconditionally (not gated on DEBUG).
+
+        Keyed by the global ``start_idx``, so rows from starts running concurrently on different
+        MPI ranks demux cleanly no matter how they interleave -- the consumer groups by start and
+        orders by iteration. Each row is a single short line written under O_APPEND, which is
+        atomic on POSIX, so concurrent multi-rank appends never tear a line. Best-effort: a write
+        failure must not abort the optimisation (the end-of-run csvs are the record of truth).
+        """
+        try:
+            path = os.path.join(self.output_dir, 'multi_start_cost_history.csv')
+            with open(path, 'a') as file:
+                file.write(f'{int(start_idx)}, {int(iteration)}, {float(cost_val):.9e}\n')
+        except OSError:
+            pass
+
     def _run_one_start(self, start_idx, x0_norm, cost_fun, gradient_func, bounds_norm):
         """Run a single bounded L-BFGS-B descent. Returns a result dict (never raises for a
         cost-converged early stop)."""
@@ -1180,6 +1200,8 @@ class MultiStartSciPyMinimizeOptimiser(Optimiser):
         start_wall = time.perf_counter()
         init_cost = cost_fun(x0_norm)
         iterates.append((init_cost, np.asarray(x0_norm, dtype=float).copy()))
+        # iteration 0 = the start point, so the live curve begins at the pre-descent cost.
+        self._append_start_cost(start_idx, 0, init_cost)
 
         last_iterate = {"x_norm": None, "cost": None}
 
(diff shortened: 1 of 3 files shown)


FinbarArgus:
Pushed a fix for the CI crash: output_dir is set on rank 0 only, so the per-rank stream hit `os.path.join(None, ...)` on rank 1+. run() now broadcasts the path to every rank (dir creation stays on rank 0); verified end-to-end at 2 ranks under the exact CI condition. Slipped through because the test suite runs at a single rank.

Issue #293, opened 2026-07-22
Laplace approximation gives massively inflated uncertainties (ill-conditioned Hessian in raw parameter space)


FinbarArgus:
## Summary

The Laplace approximation (`ia_options.method: Laplace`) produces **massively inflated parameter uncertainties** on the 3compartment model (and any model whose parameters span a wide magnitude range). The covariance it writes to `{prefix}_laplace_covariance.npy` is numerically meaningless. The code already carries `# TODO ... not correct yet. Finbar to fix.` (`calculate_hessian`) and `# TODO fix hessian calculation` (`run_laplace_approximation`).

**Interim mitigation (separate small PR):** `Laplace` now raises `NotImplementedError` pointing here, so it can't silently emit a wrong covariance until this is fixed. `profile_likelihood` is unaffected.

## Root cause

Laplace covariance is `inv(-Hessian)` of the log-posterior at the best fit (`identifiabilityAnalysis.run_laplace_approximation`). A near-singular / ill-conditioned Hessian inverts to a blown-up covariance.

The default `sub_method` is **`parabola_fit`**, whose Hessian comes from `extract_hessian_from_samples` (`utilities/utility_funcs.py`): it fits a **degree-2 polynomial** (`PolynomialFeatures` + `LinearRegression`) to sampled losses in **raw, un-normalised, un-centered parameter space**.

For 3compartment the calibrated parameters span **~17 orders of magnitude** — e.g. `aortic_root/C ≈ 1e-9` vs `E_lv_A ≈ 1e8`. The quadratic design-matrix columns (`x_i · x_j`) then span ~`1e34`, so the least-squares fit is astronomically ill-conditioned; the recovered curvature coefficients (→ Hessian) are numerical noise, and `inv(-Hessian)` explodes.

Contributing details:
- The sampling *radius* is relative (`±0.1%` of each value — fine), but the polynomial **fit** is done in absolute coordinates with **no centering on the best-fit point and no normalisation**, which is what destroys the conditioning.
- `latin_hypercube_sample_and_evaluate` already accepts a `param_norm_obj` but `calculate_hessian` doesn't pass it.
- `sub_method: numdifftools_finite_diff` is more robust (adaptive per-dimension steps) but still computes the Hessian in raw parameter space, so it's fragile for the same reason.

## Also missing: value-level tests

[...]

PR #294, opened 2026-07-22
fix(identifiability): temporarily disable the Laplace approximation (issue #293)

FinbarArgus:
Temporarily disables the Laplace approximation so it can't silently emit wrong uncertainties. **Interim mitigation for #293** (not the fix).

## Why

Laplace returns **massively inflated, numerically meaningless** parameter uncertainties for models whose parameters span a wide magnitude range (3compartment, ~`1e-9` to ~`1e8`): the covariance is `inv(-Hessian)` of a Hessian fitted (`parabola_fit`) in **raw, un-normalised parameter space**, which is astronomically ill-conditioned. The code already carries `# TODO ... not correct yet. Finbar to fix.`. Full diagnosis and the real fix (normalise the Hessian, Jacobian-transform the covariance back, add value-level tests) are in **#293**.

The existing tests are smoke-only (covariance file exists / square / finite), so a covariance full of astronomically large numbers passed — the bug was uncaught.

## What

- `IdentifiabilityAnalysis.run` raises `NotImplementedError` (pointing at #293) for `method: Laplace`, on **every rank** so an MPI run fails together. The original rank-0 dispatch is preserved in a comment to restore when #293 lands. `profile_likelihood` is unaffected; Laplace stays a valid (if disabled) `ia_options` choice.
- The **OMEX pipeline** auto-includes identifiability, and a hard failure there would take down the whole pipeline (SA + calibration already succeeded). So the generated pipeline now **catches** the `NotImplementedError`, skips identifiability, and records `laplace_*_path = null` in `pipeline_summary.json` instead of crashing.

## Tests

- `test_identifiability_python_user_defined_laplace_disabled` (renamed from `..._succeeds`): the run reaches identifiability after calibration and raises `NotImplementedError` with `#293`; no covariance is written. **Passes.**
- `test_omex_analysis_pipeline`: asserts the pipeline still succeeds and records `laplace_*_path = None`. Both tests carry a note on how to restore them when #293 is fixed.
- `test_mcmc_and_laplace_require_is_mle_cost` and the `ia_options` schema test still pass.

## Verification note

The identifiability disablement test passes here. The full OMEX pipeline test is **skipped in my environment** (its venv Python has no CellML backend — a pre-existing skip), so it'll first exercise in CI. I verified the generated pipeline script **compiles** and the skip logic is correct statically, and confirmed the dispatch raises on all ranks — but the end-to-end OMEX subprocess run is unverified here.

### DIFF
--- a/src/identifiabilty_analysis/identifiabilityAnalysis.py
+++ b/src/identifiabilty_analysis/identifiabilityAnalysis.py
@@ -131,9 +131,21 @@ class IdentifiabilityAnalysis():
         if ia_options['method'] == 'profile_likelihood':
             self.run_profile_likelihood(ia_options)
         elif ia_options['method'] == 'Laplace':
-            if self.rank == 0:
-                # currently Laplace is not parallelised, so only run on rank 0
-                self.run_laplace_approximation(ia_options)
+            # TEMPORARILY DISABLED (issue #293). The Laplace covariance is computed by inverting a
+            # Hessian fitted in raw, un-normalised parameter space; for models whose parameters
+            # span a wide magnitude range (e.g. 3compartment, ~1e-9 to ~1e8) that fit is
+            # astronomically ill-conditioned, so the uncertainties come out massively inflated and
+            # numerically meaningless. Raise instead of emitting a wrong covariance. Raised on
+            # every rank so an MPI run fails together, not just rank 0. When #293 is fixed, restore
+            # the original dispatch below (which the raise replaced):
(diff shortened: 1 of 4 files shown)


FinbarArgus:
Closing in favour of a condition-based approach: instead of blanket-disabling Laplace, the replacement PR checks the Hessian's condition number and only errors when it is ill-conditioned, and adds a selectable gradient source (AD/FSA/FD) for the Hessian, exposed in the schema for CUFLynx.

PR #295, opened 2026-07-22
feat(identifiability): usable Laplace via Jacobi-normalised Hessian, with condition check + selectable AD/FSA gradient

### LINKED ISSUE #293: Laplace approximation gives massively inflated uncertainties (ill-conditioned Hessian in raw parameter space)

FinbarArgus:
Makes the Laplace approximation usable for models whose parameters span a wide magnitude range (3compartment, ~`1e-9`..`1e8`), and replaces the temporary blanket disable (closed #294) with a **normalised condition check** plus a **selectable gradient source** exposed in the schema for CUFLynx. Closes #293.

## Why

Laplace built and inverted the precision (Hessian / Fisher information) in **raw** parameter space. For wide-magnitude models that is broken twice over: the finite-difference Hessian itself comes out as indefinite garbage (all-negative diagonal), and even a good Hessian is astronomically ill-conditioned to invert *purely from the scaling*. Either way the covariance was **massively inflated, meaningless uncertainties** (silently, with a pseudo-inverse masking it).

## What

**1a. Compute the FD Hessian in normalised space.** `parabola_fit` used to sample `±0.1%` of each parameter's *value* and fit the quadratic in raw space — for a `~1e8` elastance that samples a vanishingly small slice of its range (curvature = solver noise), and the raw-space degree-2 design matrix spans `~1e32` (round-off). It now samples a fixed fraction of each parameter's *range* and fits in scale-normalised, best-fit-centred coordinates, transforming the Hessian back via `H[i,j] = H_z[i,j]/(scale_i·scale_j)`. The recovered precision now has a solidly positive diagonal (every curvature is O(1) in normalised space). Samples are clipped to the parameter bounds so a near-bound best fit can't draw an out-of-prior (`-inf`) sample.

**1b. Invert in Jacobi-normalised space.** `run_laplace_approximation` rescales the precision to unit diagonal `P_norm[i,j] = P[i,j]/√(P_ii·P_jj)`, inverts, transforms back `C[i,j] = C_norm[i,j]/√(P_ii·P_jj)`. `C == inv(P)` exactly, so the answer is unchanged — this only removes the *scaling* part of the ill-conditioning so the intermediate inverse is numerically accurate. **Wide-magnitude-but-identifiable models now get a real covariance.**

**2. Condition guard on the normalised matrix.** `np.linalg.cond` is measured on `P_norm`, where it reflects genuine parameter **correlation** rather than magnitude spread. Above `1e12` it raises a clear `RuntimeError` — now meaning the parameters are *genuinely (jointly) non-identifiable*, not merely differently scaled. A parameter with non-positive curvature (no information) raises a distinct error. The old masking pseudo-inverse is gone.

**3. Selectable gradient source, in the schema.** `ia_options.gradient_source` (`FD` | `AD` | `FSA`, default `FD`) added to `ANALYSIS_OPTIONS` so CUFLynx discovers it. `AD`/`FSA` build the Fisher information matrix `Jᵀ diag(1/σ²) J` from the analytic observable sensitivities (the sources `gradient_sources(model_type, solver)` advertises, PR #292); `FD` uses the finite-difference `sub_method`; an unavailable source raises with the available set.

**4. OMEX pipeline resilience.** The generated pipeline catches a Laplace error and skips identifiability (recording `null`) rather than failing the whole pipeline after SA + calibration already succeeded.

## Tests

`tests/test_laplace_identifiability.py` (mock/analytic, no model): FIM assembly `== Jᵀ W J`; a wide-magnitude *identifiable* FIM (raw cond `1e32`) inverts to the correct diagonal covariance (fails before the inversion fix); genuinely correlated params still **raise**; an information-less param **raises**; the normalised Hessian fit recovers a known Hessian spanning `1e-16..1e16` exactly where the raw fit cannot; `_param_fit_scale` fallbacks.

Integration: `3compartment` (plain + `extra_ops`), `simple_physiological`, `test_fft` (finite covariance) and the python user-defined identifiability test all pass; the real-model `Simple_ODE_Benchmark` FD Laplace still validates to `diag(0.01, 0.09)`, positive definite; the `numdifftools_finite_diff` path is untouched. The OMEX test accepts Laplace succeeding or being skipped.

## For CUFLynx

`ia_options.gradient_source` is in the schema — offer only `gradient_sources(model_type, solver)` (PR #292) values plus `FD`. `AD`/`FSA` need a compatible model_type/solver + `do_ad`. A genuinely non-identifiable result surfaces as a `RuntimeError` the UI can show as "parameters not jointly identifiable."

### DIFF
--- a/src/identifiabilty_analysis/identifiabilityAnalysis.py
+++ b/src/identifiabilty_analysis/identifiabilityAnalysis.py
@@ -196,37 +196,43 @@ class IdentifiabilityAnalysis():
 
     def _invert_precision_normalised(self, precision, gradient_source):
         """Invert the precision (Hessian) matrix to a parameter covariance, using symmetric
-        (Jacobi) diagonal normalisation so that a wide spread of *parameter magnitudes* does not,
-        by itself, make the matrix look non-invertible.
+        diagonal normalisation so that a wide spread of *parameter magnitudes* does not, by itself,
+        make the matrix look non-invertible.
 
-        The precision ``P`` is rescaled to unit diagonal, ``P_norm[i,j] = P[i,j]/sqrt(P_ii P_jj)``,
-        before inverting; the covariance is transformed straight back,
-        ``C[i,j] = C_norm[i,j]/sqrt(P_ii P_jj)``. This is a mathematical identity -- ``C`` equals
-        ``inv(P)`` exactly (in exact arithmetic) -- so it does not change the answer, it only
+        The precision ``P`` is rescaled by ``1/sqrt(|P_ii|)`` on each side (unit-magnitude
+        diagonal) before inverting; the covariance is transformed straight back,
+        ``C[i,j] = C_norm[i,j]/sqrt(|P_ii| |P_jj|)``. This is a mathematical identity -- ``C``
+        equals ``inv(P)`` exactly (in exact arithmetic) -- so it does not change the answer, it only
         removes the *scaling* part of the ill-conditioning. The condition number is then taken on
         the normalised matrix, where it reflects genuine parameter *correlations* rather than mere
         magnitude spread. Models whose parameters merely span a wide magnitude range (e.g.
         3compartment, ~1e-9..1e8) therefore succeed with a real covariance, while genuinely
         collinear (non-identifiable) parameters still trip the condition guard. Issue #293.
 
+        ``abs`` (not ``sqrt(P_ii)``) is used so a *mildly indefinite* finite-difference Hessian --
+        a slightly negative curvature from FD/optimiser noise at a rough optimum -- still yields a
+        finite covariance (as the plain inverse always did) rather than aborting the whole run; a
+        non-positive-definite result is warned about, not fatal. Only a parameter with *zero*
+        curvature (literally no information, infinite variance) is a genuine abort.
+
         Returns ``(covariance_matrix, cond)`` where ``cond`` is the normalised precision's
-        condition number. Raises ``RuntimeError`` if a diagonal entry is non-positive (a flat,
-        information-less parameter direction) or if the normalised matrix is still ill-conditioned.
+        condition number. Raises ``RuntimeError`` for a zero/non-finite-curvature parameter or if
+        the normalised matrix is still ill-conditioned (jointly non-identifiable parameters).
         """
         precision = np.asarray(precision, dtype=float)
         diag = np.diag(precision)
-        # A non-positive (or non-finite) curvature on the diagonal means that parameter is flat
-        # -- it carries no information at the best fit (or we are not at a minimum) -- so its
-        # variance is undefined and Jacobi scaling would divide by zero. Surface it clearly.
-        if not np.all(np.isfinite(diag)) or np.any(diag <= 0):
-            bad = [i for i, d in enumerate(diag) if not (np.isfinite(d) and d > 0)]
+        # Only a *zero* (or non-finite) curvature is fatal: that parameter carries no information at
+        # all, so its variance is genuinely undefined (infinite) and the normalisation would divide
+        # by zero. A merely negative curvature is kept -- see the abs() below.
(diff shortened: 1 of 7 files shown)


FinbarArgus:
@MOH-Shafizadegan can you check this and confirm you are happy with it being merged?

MOH-Shafizadegan:
I've reviewed the code and the changes look good to merge

Issue #122, opened 2025-11-19
All of the parsing of information to get e.g. the param_id_info, obs_info dicts in param_id.py should be in parser functions


FinbarArgus:
This would allow sensitivity analysis to call the same functions. Currently multiple functions have been rewritten in sobolSA.py, which is bad practice. 

@MOH-Shafizadegan this would allow the sobolSA class to be much more concise and would remove rewritting of functions.

FinbarArgus:
Largely delivered by the ObsAndParamDataParser refactors and the shared param-id engine accessors — sensitivity analysis now builds its inputs through the same parser/engine functions as calibration rather than re-implementing them in sobolSA. Closing; please file focused issues for any specific parsing still duplicated.

PR #317, opened 2026-07-26
feat: AADC stiff ODE support — bdf_tape + bdf_kernel (6x faster than CasADi)

lakshtanov:
## Summary

Add two new solver methods for stiff ODE calibration with AADC:

- **bdf_tape**: records entire BDF integration on AADC tape (Python). First call ~1.5 min recording (with jac_lag), subsequent ~2s eval.
- **bdf_kernel**: C++ kernel replay via ConstStateExtFunc. First call ~6s, subsequent **0.3s eval** (cached tape).

### Benchmark (3compartment, 27 states, 4 params, pre_time=20, sim_time=2)

| Method | Total optimizer time |
|--------|---------------------|
| **AADC bdf_kernel** | **26s** |
| CasADi BDF | 153s |
| AADC bdf_tape | ~3 min |

### Gradient verification

AD/FD = 1.0000 for all RHS parameters (central FD, h=1e-6).

### What is included

- bdf_tape: semi-implicit BDF on tape, diagonal FD Jacobian, jac_lag optimization, all 6 observables (state + var via compute_variables)
- bdf_kernel: C++ kernel replay with ConstStateExtFunc (forward + reverse AD), tape caching. Falls back to bdf_tape if C++ module not available.
- 5 gradient bug fixes from previous commits (indentation, warmup sensitivity, BDF2 formula, algebraic var gradient, initial state params)

### DIFF
--- a/benchmarks/aadc_3compartment_results.md
+++ b/benchmarks/aadc_3compartment_results.md
@@ -67,9 +67,30 @@ All 4 parameters verified at nominal values with central FD (h=1e-6):
 
 AADC **16-20× faster** than both FD and CasADi on non-stiff models.
 
-### Next steps to close the gap with CasADi on stiff models
+### Tape approach: semi-implicit on AADC tape (2026-07-24, proof-of-concept)
 
-1. **Skip early warmup sensitivity**: first ~90% of warmup reaches steady state; only last few cardiac cycles matter. Could save 9× on warmup dfdp evals.
-2. **Batch VFJ**: vectorize kernel replay across sub-steps (reduce Python overhead)
-3. **C++ Newton loop**: move Newton + IFT to C++ to eliminate Python per-step overhead
-4. **AADC tape for compute_variables**: replace FD chain rule with AD for var obs
+**Key insight**: record entire BDF integration as idouble operations on one AADC tape.
+Forward+reverse replay in C++ eliminates Python loop overhead.
+
+| Metric | Python IFT | **Tape semi-implicit** | CasADi |
+|--------|-----------|----------------------|--------|
+| Per eval | 46s | **2.2s** | 2s |
+| Gradient AD/FD | 1.000 | **1.000** | 1.000 |
+| Forward error | 0% | 0.97% | 0% |
+| Recording | 0 | 10 min (one-time) | 0 |
+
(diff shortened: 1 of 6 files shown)


FinbarArgus:
Looks great!

I've created this issue in CUFLynx so we can bring AADC into our GUI. Let me know your thoughts on easy ways to let a user install the AADC academic licence in the python shipped with CUFLynx? Or we might need to get them to use a separate venv python.

https://github.com/physiomelinks/CUFLynx/issues/122

NatMatLogica:
Thanks @FinbarArgus — happy to make this straightforward. The cleanest approach keeps AADC as a **user-installed dependency**, not something bundled into CUFLynx. That keeps the licence boundary clean and means no one is forced to have AADC to use CUFLynx.

**Mechanism**

- **CUFLynx does not ship AADC.** The user installs their own licensed copy into the Python environment CUFLynx uses (a dedicated venv is fine). CUFLynx — reading circulatory_autogen's schema — checks whether `aadc` is importable with a valid licence: if yes, the AADC solvers (`bdf_kernel` / `bdf_tape`) are enabled; if not, it **falls back to CasADi automatically**. PR #317 already implements that fallback, so AADC stays a fully optional accelerator.
- This keeps circulatory_autogen and CUFLynx Apache-2.0, with AADC as a separately-licensed optional add-on.

**Getting a licence key**

- AADC is available free for academic / non-commercial use. To request an academic licence key, email **Natalija Karpichina — natalija.karpichina@matlogica.com** with your name, institution, and a line on intended use, and we'll issue a key plus install instructions. Commercial users can reach the same address for a commercial licence.

**Install**

- Once you have a key, install the AADC Python package we provide and point it at your licence key (full steps come with the key). CUFLynx then detects it automatically.

FinbarArgus:
Hi @NatMatLogica,

Cool, that makes the license access clear.

The other thing I need to figure out is how to easily install an extra library in the python shipped with CUFLynx. We ship a python with all of the other libraries included, so no pip installs are required. However for AADC we will need to enable using pip install with the shipped python, maybe through a terminal window within the GUI.

FinbarArgus:
merged and closed with #340

PR #379, opened 2026-08-08
fix(solvers): Myokit set_tolerance args were swapped; CVODE_myokit defaults become rel 1e-6 / abs 1e-8

FinbarArgus:
## 1. `set_tolerance` arguments were swapped (accuracy bug)

Myokit's signature is **`set_tolerance(abs_tol, rel_tol)`** — absolute *first*, the reverse of the rtol-then-atol order CA's schema lists. `myokit_helper` passed them positionally in schema order, so each reached the other argument. Invisible while CA's declared default was a symmetric `1e-8/1e-8`; measured on 3compartment (by the CUFLynx agent), `rtol=1e-6, atol=1e-4` — the names reversed — produced a cost **bit-identical** to Myokit's own `abs 1e-6 / rel 1e-4` defaults, which is only possible if the arguments were exchanged.

The application now lives in `apply_cvodes_tolerances`, called **by keyword** so the mapping cannot silently swap again. A partially-set pair fills the missing partner with CA's declared default. The FSA behaviour is unchanged: still forces `1e-8/1e-8` when sensitivities are enabled and the user set none.

## 2. `CVODE_myokit` defaults become rel 1e-6 / abs 1e-8

Front-ends seed interactive solves from the declared schema defaults rather than restating them (CUFLynx #200), so the declaration is the one place the default is decided. The absolute tolerance **keeps the 1e-8 floor previous users ran at**, so existing models do not start failing; only the relative knob relaxes to 1e-6 — the relative tolerance is where most of the `1e-8/1e-8` interactive solve cost was (`1e-8/1e-8` measured ~2.3× slower than Myokit's own defaults on 3compartment). The helper now **applies** these defaults whenever the user sets neither value, so declared and effective cannot drift whichever front door the run came through; explicit values and the FSA `1e-8/1e-8` tightening still win. Only `CVODE_myokit` changes; the rest of the CVODE family keeps `1e-8/1e-8`.

## Tests

All new tests use **asymmetric** pairs (a symmetric pair cannot catch the swap) and assert which value reached which keyword argument:

- rtol→`rel_tol` / atol→`abs_tol` mapping (fails against the positional call)
- no tolerances → CA's declared defaults applied (abs 1e-8, rel 1e-6), pinned equal to the schema declaration
- FSA tightening to 1e-8/1e-8, and explicit tolerances winning over it
- partial setting fills the partner with CA's default
- a failed solve warns with the effective MaximumStep / atol / rtol and that decreasing them may help stability
- `inspect.signature` check that Myokit still takes `abs_tol`/`rel_tol` keywords — a rename breaks loudly
- schema defaults pinned to the helper's constants

Reported by the CUFLynx agent (`handover_CA_tolerances.md`).

### DIFF
--- a/src/parsers/PrimitiveParsers.py
+++ b/src/parsers/PrimitiveParsers.py
@@ -259,7 +259,16 @@ _CVODE_FAMILY_SOLVER_INFO = [
 _MYOKIT_SOLVER_INFO = [
     {'name': 'MaximumStep', 'type': 'float', 'default': 0.001, 'required': False,
      'description': 'Maximum integrator step size (myokit Simulation.set_max_step_size).'},
-    _SI_RTOL, _SI_ATOL,
+    # Myokit's own set_tolerance defaults (rel 1e-4, abs 1e-6), not the 1e-8/1e-8 the rest of
+    # the CVODE family declares. Front-ends seed interactive solves from these declared
+    # defaults, and 1e-8/1e-8 costs ~2.3x per solve on 3compartment (209 ms vs 92 ms) with no
+    # accuracy need on the plain forward path -- tighter is a deliberate user choice. FSA is
+    # unaffected: it still forces 1e-8/1e-8 when the user set none, because a sloppy forward
+    # solve makes a poor gradient (see myokit_helper.apply_cvodes_tolerances).
+    {**_SI_RTOL, 'default': 1e-4,
+     'description': "Relative integration tolerance (default: Myokit's own, 1e-4)."},
+    {**_SI_ATOL, 'default': 1e-6,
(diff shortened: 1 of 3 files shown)


FinbarArgus:
Added in dba5126 (requested by Finbar): a failed Myokit solve now emits a warning naming the three stability knobs with their **effective** values — `MaximumStep is X, atol is X, rtol is X; decreasing these (solver_info MaximumStep / atol / rtol) may help stability`. `apply_cvodes_tolerances` returns the effective `(abs_tol, rel_tol)` pair (Myokit defaults included) so the diagnostic reports what the integrator actually ran with; an unset MaximumStep is reported as `unset (unbounded)` rather than an invented number. Unit tests cover the hint text and both MaximumStep cases.

PR #380, opened 2026-08-08
feat(params): FSA sensitivities follow the grouped/modifier chain rule; gradient_method speaks AD/FSA

FinbarArgus:
## What

Since #376, `set_param_vals` moves **every** member of a grouped `params_for_id` row, but both local-sensitivity arms flattened a group to its **first member**. On the Myokit/FSA path the cost was therefore a function of all members while the gradient tracked only the first — the exact cost-vs-gradient mismatch #376 refused to introduce for CasADi, live and silent on FSA.

This PR makes the FSA arm exact for all three cases:

```
ungrouped            dJ/dp
shared-value group   dJ/dtheta = sum_i  dJ/dp_i
scale modifier       dJ/dtheta = sum_i (dJ/dp_i) * baseline_i
```

### How

- `ensure_setup` hands **every member qname** to `enable_fsa` (a CVODES column per member) and records a per-entry `(member, weight)` list — weight 1 for shared-value groups, the resolved baseline for scale modifiers (`modifier_weights_by_index`).
- Perturbing theta by `h` moves member *i* by `w_i * h`, so the directional derivative runs along the **combined** sensitivity `sum_i w_i * S_i` (`combined_entry_sensitivities`). The perturbation machinery itself is unchanged; the three cases fall out of the weights. Both `get_jac_cost` and `observable_feature_sensitivities` use the combination.
- An entry with any FSA-ineligible member falls back to full-cost FD **as a whole** (theta FD is exact for groups since the cost path expands theta).
- The modifier refusal (`_refuse_sensitivities_for_modifiers`) now guards only the CasADi arm, which still cannot express a grouped row (`_create_param_subset`); symbolic substitution there is follow-up work.

### Reporting keyed by entry label

Sensitivity dicts are now keyed by `param_entry_labels` (new accessor; `param_labels` with a fallback for older dicts): a grouped entry's number is d/dtheta over all members, and keying it by one member's qname reported it as a different quantity. Both consumers (local SA matrix, Laplace FIM) and the multi-start CSV headers read the same labels. Ungrouped rows are unchanged (label == qname).

### Modifier fixes surfaced by a real calibration

Driving a modifier through an actual 3compartment calibration (below) surfaced three latent #378 gaps, all fixed here:

1. **Ragged x0**: `get_init_param_vals` returns a *list* per multi-name entry, and `np.asarray` over that crashed every optimiser's x0 handling (`sp_minimize` + any grouped row was broken since #376). `param_init` is now flat, one slot per theta.
2. **Modifier x0/nominal semantics**: theta's start (and local-SA nominal) is now the operation's **identity** (`scale` → 1.0, new introspectable `identity` field in `PARAM_MODIFIER_OPERATIONS`), not the first target's model default — which would have scaled every target by ~1e-8.
3. **`param_modifiers.json` saved with `baselines: null`**: the parse-time save runs before a simulation helper exists. The calibration run now re-saves once baselines are resolved (`save_param_modifiers`, idempotent).

### gradient_method speaks the arm names (CUFLynx ask)

`get_observable_sensitivities` now accepts `'AD'` and `'FSA'` alongside `'FD'`/`'auto'`/`'analytic'` — the same vocabulary as `gradient_sources()` and the Laplace `gradient_source` — validating each against the backend rather than silently reinterpreting: `'AD'` off `casadi_python` raises naming the mismatch; `'FSA'` without `cellml_only` + `CVODE_myokit` + `do_ad` raises naming exactly what is missing. The `sa_options` schema advertises `['auto', 'AD', 'FSA', 'FD']` (default `auto`); `'analytic'` stays accepted in code as a legacy spelling but is no longer advertised. (Requested in `handover_CA_gradient_method_names.md`; included here because it edits the same dispatch the chain-rule work restructured.)

## Tests

- `tests/test_fsa_grouped_sensitivities.py` — unit tests that **discriminate by construction** (synthetic member sensitivities of comparable magnitude): the flattened member list reaching `enable_fsa`, plain-sum vs baseline-weighted-sum gradients, first-member-alone asserted unreachable, whole-entry FD fallback, label keying, identity nominals, refusal placement. An FSA-vs-FD test on a real model cannot catch this bug (measured: wrong and right answers agree to 1e-5 relative on 3compartment `[aortic_root/C, par/C]`).
- `tests/test_param_id.py::test_param_id_3compartment_modifier_calibration_fsa` — the real-calibration modifier run (#378 shipped unit tests only): theta scaling both compliances through `sp_minimize` + `do_ad` on `CVODE_myokit`, asserting `param_modifiers.json` lands with the model-default baselines and theta stays a dimensionless value in bounds.
- `tests/test_fd_observable_sensitivities.py` — each gradient_method spelling reaches the arm it names; mismatched pairs raise naming the mismatch/missing pieces.
- All unit tests verified failing against the pre-fix code (the buggy gradient returns exactly the first member's derivative).
- Full suite run: 599 passed, 7 failed — the six failures already present on `upstream/master` (SN_simple python codegen, AADC 35.1% deviation ×3, AADC untapeable DID-NOT-RAISE) plus one `inspect.getsource` artifact from editing sources while the suite ran, verified passing standalone.

### DIFF
--- a/src/identifiabilty_analysis/identifiabilityAnalysis.py
+++ b/src/identifiabilty_analysis/identifiabilityAnalysis.py
@@ -172,8 +172,10 @@ class IdentifiabilityAnalysis():
                 "'FD' for finite differences.)")
 
         sens = pid.get_observable_sensitivities(np.asarray(self.best_param_vals, dtype=float))
-        param_names = [n[0] if isinstance(n, (list, tuple)) else n
-                       for n in pid.param_id_info['param_names']]
+        # The arms key their result dicts by entry label (one per calibrated theta, matching
+        # best_param_vals); first-member qnames would miss every grouped column and read 0.0.
+        from parsers.PrimitiveParsers import param_entry_labels
+        param_names = param_entry_labels(pid.param_id_info)
         obs_info = pid.obs_info
         const_to_obs = obs_info['const_idx_to_obs_idx']
         stds = np.asarray(obs_info['std_const_vec'], dtype=float)
--- a/src/param_id/casadi_backend.py
+++ b/src/param_id/casadi_backend.py
@@ -200,8 +200,12 @@ def get_observable_sensitivities(pid, param_vals):
     output instead of the aggregated cost.
     """
     import casadi as ca
+    from parsers.PrimitiveParsers import param_entry_labels
     param_names = pid.param_id_info["param_names"]
-    flat_names = [n[0] if isinstance(n, list) else n for n in param_names]
+    # Grouped rows never reach here (_create_param_subset refuses them), so each label is its
+    # entry's single qname today -- but keying by param_entry_labels keeps the report shape
+    # identical to the Myokit FSA arm, which does handle groups and modifiers.
+    flat_names = param_entry_labels(pid.param_id_info)
     build_functions(pid, param_names, param_vals)
 
     # jacobian of the flattened observable vector w.r.t. the differentiated parameter subset.
--- a/src/param_id/fd_backend.py
+++ b/src/param_id/fd_backend.py
@@ -21,6 +21,8 @@ always something they chose.
 """
 import numpy as np
 
+from parsers.PrimitiveParsers import param_entry_labels
+
 
 def _step(pj, pmin, pmax, h):
     """The central-difference step for one parameter.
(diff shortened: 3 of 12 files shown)


FinbarArgus:
Verification summary: the full suite (599 passed / 7 failed) showed only the six failures already present on upstream/master plus one `inspect.getsource` artifact from editing sources while the suite ran (passes standalone). A targeted re-run of the affected files against the final code then caught one real regression from the flat-x0 change — CMA-ES's out-of-bounds x0 warning did a bare truthiness check on what is now an ndarray — fixed in b63b611 and confirmed passing. The real-calibration modifier test (`test_param_id_3compartment_modifier_calibration_fsa`) passes end to end on the FSA chain-rule path.

Issue #346, opened 2026-08-03
aadc_semi_implicit: methods_by_solver advertises gradient strategies as integrators, and the forward dispatch rejects two of them


FinbarArgus:
Found while measuring which AADC methods can integrate `3compartment` (downstream report: physiomelinks/CUFLynx#175, fixes physiomelinks/CUFLynx#176).

`SOLVER_SCHEMA['methods_by_solver']['aadc_semi_implicit']` is what a GUI builds its method dropdown from — CUFLynx reads it directly, on the principle that CA's schema is the single source of truth for these lists. Four separate things in it do not hold.

## 1. The list mixes integrators with gradient strategies

```python
'aadc_semi_implicit': ['adaptive_rk45', 'semi_implicit', 'implicit_euler_ift',
                       'implicit_newton', 'bdf_newton', 'bdf_tape', 'bdf_kernel', 'rk4']
```

`bdf_tape` and `bdf_kernel` are not integrators. They exist only in `param_id/aadc_backend.py`, as `_cost_and_grad_bdf_tape` / `_cost_and_grad_bdf_kernel`, and do their own stepping — there is no branch for either in `aadc_python_solver_helper.run()`. So a plain forward solve with one selected raises:

```
ValueError: Unknown AADC solver_info method 'bdf_tape'.
```

Because one `solver_info['method']` serves both the forward solve and the gradient, a user who picks `bdf_tape` for calibration breaks every interactive simulation, and a user who wants an interactive simulation cannot pick it at all.

**Ask:** publish the two sets separately — e.g. `forward_methods_by_solver` alongside the existing key, or a `kind` on each entry. A consumer cannot currently tell them apart without hardcoding the names, which is exactly what reading the schema is supposed to avoid. CUFLynx has had to hardcode them in the meantime (`solver_options.UNSUPPORTED_METHODS`), and that exclusion is documented to be retired the moment this lands.

## 2. The `ValueError` lists four methods; the dispatch accepts six

```python
# aadc_python_solver_helper.run() accepts:
implicit_euler_ift | implicit_newton | bdf_newton | semi_implicit | rk4 | adaptive_rk45

# ...and the error it raises says:
"Valid methods are 'adaptive_rk45', 'semi_implicit', 'implicit_euler_ift', 'rk4'."
```

[...]

PR #381, opened 2026-08-09
feat: restore bdf_kernel -- C++ kernel replay, 13.5x faster than scipy BDF

lakshtanov:
## Restore bdf_kernel -- C++ kernel replay (13.5x faster)

Restores the step-kernel architecture from PR #340 (lost during refactoring).

### Benchmark (3compartment, 27 states)

| Test | Result |
|------|--------|
| Cost match vs scipy BDF | 8.5e-4 (forward trajectories agree) |
| Gradient accuracy | AD/FD = **1.0000** |
| Speed | bdf_kernel **0.068s** (cost+grad) vs scipy BDF **0.921s** (forward only) = **13.5x** |

### How it works

Two-level recording: small inner kernel (ODE RHS) wrapped inside a single-step tape, replayed in C++.

1. Record compute_rates(t, y) once -> f-kernel (~1KB)
2. Wrap as KernelExtFunc (ConstStateExtFunc) for use inside another recording
3. Record ONE BDF step: wrapped f-kernel call + semi-implicit update + diagonal Jacobian (off-tape)
4. C++ loop replays step tape 22K times -- no Python boundary crossings
5. Gradient via IFT at converged steps

### Files

- src/param_id/aadc_native/bdf_loop.cpp -- C++ kernel replay (KernelExtFunc, LUSolveExtFunc)
- src/param_id/aadc_backend.py -- driver wiring aadc.bdf_record_and_evaluate into param_id
- tests/test_bdf_kernel.py -- 3 tests (cost match, gradient vs FD, speed benchmark)

bdf_record_and_evaluate is already in the AADC Python package -- no compilation needed.

### DIFF
--- a/src/param_id/aadc_backend.py
+++ b/src/param_id/aadc_backend.py
@@ -1,0 +1,991 @@
+"""AADC (Matlogica) tape-based cost and gradient for the param-id classes.
+
+The AADC backend records the forward integration on a tape and replays it, so the cost has
+to be re-implemented on that tape rather than reusing the generic cost-assembly path -- both
+the cost and its gradient then come out of one tape evaluation and cannot drift apart. That
+is why this module, unlike casadi_backend and fsa_backend, calls no generic method of the
+param-id object at all: it needs only ``pid.sim_helper`` and the raw obs/protocol/param
+dicts.
+
+``cost_and_grad(pid, param_vals)`` takes the param-id object as its first argument. Nothing
+here imports ``param_id.paramID``; the module imports standalone.
+
+AADC is optional third-party proprietary software. It is imported lazily inside
+``cost_on_tape`` (i.e. only once a tape is actually being recorded), exactly as before the
+extraction -- hoisting it to module scope would change when ImportError surfaces on a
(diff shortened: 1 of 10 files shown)


FinbarArgus:
Rebased onto current master as #382 (the branch here could not be updated in place — pushing to `lakshtanov:feat/bdf-kernel-restore` was not possible from this session, apologies). #382 keeps the bdf_kernel commit with @lakshtanov's authorship, drops `25d4698` (its content is already on master via the earlier AADC PRs), and resolves `aadc_backend.py` against master's #346 method/strategy dispatch instead of reverting it — only the kernel fixes are applied (new `bdf_record_and_evaluate` signature with `compute_variables_fn`, var-observable support, updated `bdf_loop.cpp`, the 3 new tests). One finding from running them: no released AADC wheel (1.8.0/1.8.2) exports `_aadc_core.bdf_record_and_evaluate`, so the new tests skip and the kernel strategy falls back to `bdf_tape` — @lakshtanov, is the entry point slated for an AADC release? Details in #382.

PR #385, opened 2026-08-10
feat(params): modifier functions — theta computes its targets through a registered function (#383)

### LINKED ISSUE #383: Support calibrating a new parameter which allows other parameters to be calculated from it
e.g. If we have q_tot = 5 as the parameter we want to calculate.

Have q_lv_init = q_tot - q_other_init

FinbarArgus:
Implements #383: a calibrated theta whose targets are **computed from it** through a registered, user-definable **modifier function**:

```
p_i = fn(theta, baseline_i, **inputs)
```

The motivating case works out of the box with the new built-in `remainder`:

```json
{"name": "q_tot",
 "modifies": ["heart/q_lv_init"],
 "operation": "remainder",
 "inputs": {"subtract": ["pvn/q_init", "par/q_init"]},
 "min": 4e-3, "max": 6e-3}
```

→ `q_lv_init = q_tot − (q_pvn_default + q_par_default)`. `scale` (#378) becomes the built-in `theta * baseline` — same numbers, one code path.

## Design (as agreed with Finbar)

- **`operation` names a registered function** — no breaking change to existing `scale` entries. Built-ins live in `src/param_id/modifier_funcs.py`; user functions in `funcs_user/modifier_funcs_user.py` with the `@modifier_func(inputs={name: 'float'|'list'})` decorator; external files via the new `modifier_funcs_external_path` config key (threaded like the operation/cost-func paths through `CVS0DParamID`, `SensitivityAnalysis`, `sobolSA` and `ObsAndParamDataParser`).
- **`inputs` supply model qnames; the function receives their *default* values** — `'float'` = one qname, `'list'` = several. Resolved once at setup exactly like the baselines, so nothing compounds across iterations.
- **Affine-in-theta, enforced**: every function is numerically probed at resolve time (`p = a·theta + b` for the resolved inputs). `a` is the constant chain-rule weight the FSA gradient applies per member (`modifier_weights_by_index` now returns probed `a`; for scale that is the baseline — the #380 machinery is reused unchanged, verified by an ensure_setup test). Non-affine functions are refused with a message explaining why, before they can make a gradient silently wrong.
- **theta's x0 / local-SA nominal auto-derives** by inverting the mapping at the first target's baseline — the run starts with every target at its model default. Scale inverts to exactly 1.0, so #378 behaviour is unchanged.
- **CUFLynx schema**: `param_modifier_operations(external_path)` now returns the merged registry — each record carries `description`, `applies_to`, `inputs` (what the entry must supply qnames for, and of which type) and `user_defined`; scale keeps its static UI metadata (`default_min`/`default_max`/`dimensionless`/`identity`). CUFLynx renders the modifier form from this data, per CUFLynx #200's no-restating rule.
- Entry validation is at parse time: unknown operation, missing/undeclared inputs, and float-vs-list type mismatches all fail naming the declared inputs. `param_modifiers.json` now records `inputs`, `resolved_inputs` and the probed `affine` coefficients, so a saved theta stays interpretable.

## Tests

20 new tests in `tests/test_modifier_funcs.py` (registry/decorator/external file; vocabulary shape for CUFLynx; entry validation; the q_tot case end to end including no-compounding and x0 inversion; probed weights reaching the FSA chain rule; non-affine refusal) plus the full existing modifier/grouped/JSON/FSA suites — 231 affected tests pass, and the real-calibration modifier integration test passes end to end on the FSA path.

Closes #383.


FinbarArgus:
Added the end-to-end `remainder` calibration test Finbar asked for: `test_param_id_3compartment_remainder_calibration_fsa` calibrates the total blood volume `q_tot` and derives `global/q_lv_init = q_tot − sum(all six other initial volumes)` (`global/q_ra_init`, `global/q_rv_init`, `global/q_la_init`, `pvn/q_C_init`, `systemic_T/q_init`, `venous_svc/q_C_init` — every `q*_init` in `3compartment_parameters.csv`) through `sp_minimize` + `do_ad` on `CVODE_myokit`. Passes: theta starts at the baseline total 5.868e-3 via the affine inversion, and `param_modifiers.json` records inputs, resolved defaults, and `a=1 / b=−3.868e-3`. Both built-in modifier functions now have real-calibration coverage.

FinbarArgus:
Per Finbar's review, the q_tot test now accounts for **every** constant carrying blood volume at t=0 — derived from each module's own convention rather than just the `q*_init` grep: heart chamber inits (their `q_*_us` is pressure-law-only, `E·(q − q_us)`, so adding it would double-count), `arterial_simple` reference volumes `q_0` (states start at zero, so `q(0) = q_0` — **aortic_root's 6.94e-6 was genuinely missing** from the previous list), terminal `q_init` (total-volume convention), and the venous `q_C_init` **plus** `q_us_0` (that module keeps unstressed volume outside the init; both are 0 in 3compartment but the structure is now right). One caveat recorded in the test: `q_us_0` enters the true total as `q_us_0·(1−Δq_us)`, so the constant subtraction is exact while `Delta_q_us = 0` — a model driving venoconstriction control shifts the total in a way no constant-input modifier can see. Test passes end to end (theta baseline now 5.87494e-3).

FinbarArgus:
Correction to my previous comment's caveat, after Finbar asked whether `Delta_q_us = 0` at t=0 is guaranteed: **it is not guaranteed** (constant from parameters.csv in the `vp` variant, ANS-controller-driven boundary condition in `vp_wCont`) — **but it does not matter**. The venous module's total is `q = q_C + q_us_wCont = q_C_change + q_us_0`: the `q_us_wCont` terms cancel, so venoconstriction re-partitions stressed vs unstressed volume without changing the total. `q(0) = q_C_init + q_us_0` is exact for any `Delta_q_us`, at every t. The test comment now states this instead of the earlier (wrong) 'exact while Delta_q_us = 0' caveat. There is no issue.

FinbarArgus:
Renamed for vocabulary consistency per review (a5fbb6d): **operations act on outputs (obs_data), modifiers act on parameters (params_for_id)** — #378 had borrowed `operation` for the modifier's function name.

| was | now |
|---|---|
| entry key `"operation"` | `"modifier"` |
| record field `operation` | `modifier` |
| `param_modifier_operations()` | `param_modifiers()` |
| `PARAM_MODIFIER_OPERATIONS` / `DEFAULT_PARAM_MODIFIER_OPERATION` | `PARAM_MODIFIERS` / `DEFAULT_PARAM_MODIFIER` |
| "unknown operation …" messages | "unknown modifier …" |

So an entry now reads:
```json
{"name": "q_tot", "modifies": ["heart/q_lv_init"], "modifier": "remainder",
 "inputs": {"subtract": ["pvn/q_init", "par/q_init"]}, "min": 4e-3, "max": 6e-3}
```

Back-compat kept where it matters: old module-level names are aliases (`param_modifier_operations` is the same callable), the `operation` entry key still works but warns — and is refused if set alongside `modifier` rather than silently picking one — and `_modifier_name()` reads the legacy record field so a `param_modifiers.json` from a #378-era run still resolves (tested). @CUFLynx: read `param_modifiers(external_path)` and emit `"modifier"`; the old spellings keep working meanwhile.

235 unit tests + both modifier integration tests pass.

Issue #151, opened 2026-01-27
libCellML is currently not analysing the models as ODEs (only important for Python generation)


FinbarArgus:
We need to figure out why libcellml complains about some of these models even though they can create models to be used with OpenCOR succesfully.

PR #397, opened 2026-08-12
fix(parsers): derive the solver->consumer map (#330) and read parameters.csv by header (#159)

### LINKED ISSUE #159: Ensure CSV readers are using headers
It appears that the parameter csv files are being processed based on index and not their headers. If this is the case, we should transition to using header-based parsing as it's more robust.

### LINKED ISSUE #330: Schema-vs-code check is per-setting, not per-solver, and counts pass-through as "read"
`tests/test_solver_info_validation.py::test_schema_settings_are_actually_read_by_the_code` exists to stop the schemas advertising a setting no code consumes — because downstream tools (CUFLynx) build their settings forms from `SOLVER_INFO_FIELDS`, so a phantom setting becomes a control the user can change with no effect and no way to tell.

It has two gaps that let exactly that happen (see #329):

**1. It checks per *setting*, not per *solver*.** It collects the names across all of `SOLVER_INFO_FIELDS` and greps `src/` for each. So a setting that is genuinely read by *one* backend passes for *every* solver that advertises it. `MaximumNumberOfSteps` is read by the opencor/cpp path, which was enough to clear it for `CVODE_myokit` too — where `myokit_helper` never reads it, because `myokit.Simulation` has no max-step-count knob at all.

**2. A name that merely passes through counts as read.** The grep is `'<name>' in corpus`. `protocol_runner.py` has:

```python
self.MaximumNumberOfSteps = solver_info.get('MaximumNumberOfSteps', 5000)
...
full_solver_info.setdefault('MaximumNumberOfSteps', self.MaximumNumberOfSteps)
```

which only relays the value into `get_simulation_helper`'s `solver_info` — the backend it reaches may ignore it entirely. The test's docstring cited this relay as evidence the setting was read; #329 replaces that with a recorded limitation.

## What a real fix needs

Per-solver, consumption-aware checking. Roughly:

- for each `(solver, field)` in `SOLVER_INFO_FIELDS`, establish that *that solver's* execution path consumes the field — not merely that the name appears somewhere in `src/`;
- distinguish **consuming** a key (passing it to an integrator API, branching on it, arithmetic) from **relaying** it (reading it out of one `solver_info` and putting it into another);
- keep it maintainable — a hand-maintained solver→reader-module map would drift, which is the failure mode this test was written to prevent in the first place.

One workable shape: map each solver to its helper module(s) via `solver_wrappers.get_simulation_helper`'s own dispatch (so the mapping is derived, not hand-written), then require each advertised field to appear in that module — with an explicit, commented allow-list for the genuine cross-module cases.

## Not urgent, but not cosmetic

The check is currently *necessary but not sufficient*, and it reads as sufficient. Anyone adding a backend that shares a field list with an existing one inherits the same blind spot. Worth doing before the next solver family is added.

FinbarArgus:
Closes #330. Closes #159.

Two independent fixes, both in the parsers, kept in one PR because they are small and touch adjacent files.

## #330 — the schema-vs-code check was still not per-solver in the way the issue asked

The per-solver check landed already, but with the two gaps the issue named:

**1. The solver → backend-module map was hand-written.** That is the failure mode the check exists to prevent: it drifts silently, and a solver whose entry goes stale gets checked against the wrong file — or, if the entry is simply missing, is not checked at all. It is now **derived** by reading `get_simulation_helper`'s own `if/elif` dispatch (AST walk resolving both `solver == 'x'` and `solver in some_list`), so adding a backend to the factory wires it into the check for free.

Two documented additions remain, each with a comment saying why the derivation cannot see it:
- `aadc_semi_implicit` → `param_id/aadc_backend.py` (gradient knobs are read by the tape/kernel paths, not the forward-solve helper)
- the cpp solvers → the generate script + `CVSCppGenerator` (no Python backend exists to dispatch to)

**2. "Read" was a substring search over the file text**, so a key named only in a docstring, a comment or an error message vouched for itself — `myokit_helper`'s own docstring lists supported `solver_info` keys. It is now an AST walk counting only genuine setting-name positions: subscript, `.get`/`.pop`, membership, an equality branch (the translating forwarders, `rtol` → `RelativeTolerance`), and key allow-lists / default blocks.

### Closing the cpp exemption surfaced three phantom settings

The cpp solvers were previously exempt, and that is exactly how they came to advertise settings nothing reads. The emitted C++ steps at `dt_solver` and **hardcodes its tolerances**:

```c
sunrealtype reltol = 1e-7; // TODO get this from user_inputs.yaml too
sunrealtype abstol = 1e-9; // TODO get this from user_inputs.yaml too
```

So `MaximumStep`, `rtol` and `atol` reached nothing while CUFLynx rendered a control for each. `CVODE`/`RK4`/`PETSC` now advertise only `MaximumNumberOfSteps` — the one integrator key the generate script really forwards, as CVODE's `mxsteps`. They go back the day the generator reads them.

Existing configs are **migrated with a warning naming what stopped applying, not rejected** — the same treatment #329 gave `CVODE_myokit`:

```
WARNING: solver_info key 'MaximumStep' is not supported by solver 'RK4' and was ignored.
The generated C++ integrates at a fixed step; use the framework key 'dt_solver' to set it.
```

### New guards

- the derivation must still recognise the dispatch, or every per-solver check would pass vacuously
- every advertised solver must have a consumer to check against (no silent exemptions)
- a relay may never be treated as a consumer
- a setting named only in prose does not count as read

## #159 — parameters.csv was read by column position

`__reduce_parameters_array` flattened each row of the structured array into a positional list:

```python
new_entry = [item for item in new_entry]
if new_entry[1] != param_tuple[1]:      # "units"
new_entry.insert(2, param_tuple[2])
parameters_array['value'] = np.array(parameters_list)[:, 3]
```

The fields of that array are the CSV's own headers **in the CSV's own order**, so any extra column shifted every field to its right.

`resources/FTU_wCVS_parameters.csv` is a shipped file with a `comp_env` column between `units` and `value`. Its parameters parsed as:

| field | parsed | should be |
|---|---|---|
| value | `heart` | `1050` |
| data_reference | `1050` | `known` |

Nothing raised — the row still had a plausible shape, so a model would be built from component-name strings in place of numbers. A column count that ended up inconsistent instead hit the generic `parameters rows are of non consistent length` exit, which does not say why.

Rows are now addressed by name, so **column order stops being part of the file format** and extra columns are ignored. Missing required columns are named in the error rather than becoming `EMPTY_MUST_BE_FILLED` or an `IndexError`, and `get_data_as_dataframe` strips header names as well as values so `variable_name, units` in a header row still resolves under the name the user can see.

The bare `except:` around the row lookup is narrowed to the empty-match case it was really guarding — that catch-all is what swallowed the units-mismatch `exit()` and forced the `actually_exit` re-raise dance.

## Tests

Six new tests in `tests/test_issue_fixes.py` for #159; three of them fail before the fix (extra column corrupts the value, reordered columns exit, missing column raises `IndexError`) and three pin unchanged behaviour (missing row → `EMPTY_MUST_BE_FILLED`, unit mismatch → exit, header whitespace tolerated). Four new tests in `tests/test_solver_info_validation.py` for #330, plus parametrised cpp migration/validation coverage.

Verified `not slow and not compare_optimisers`: 587 passed, 1 failed. The failure is `test_python_BDF_solver[SN_simple]`, which is the known SN-under-libCellML-Analyser limitation (#151, recorded in CLAUDE.md) and fails identically on this branch with the source changes stashed.

### DIFF
--- a/src/parsers/ModelParsers.py
+++ b/src/parsers/ModelParsers.py
@@ -17,6 +17,12 @@ import os
 generator_resources_dir_path = os.path.join(os.path.dirname(__file__), '../generators/resources')
 base_dir = os.path.join(os.path.dirname(__file__), '../..')
 
+# The columns a {prefix}_parameters.csv must provide. They are looked up by header name, so a file
+# may list them in any order and may carry extra columns (FTU_wCVS_parameters.csv has a 'comp_env'
+# column, for instance) without shifting the ones that matter -- see issue #159. 'const_type' is
+# deliberately absent: it is supplied by the module config, not by the CSV.
+_REQUIRED_PARAMETER_COLUMNS = ('variable_name', 'units', 'value', 'data_reference')
+
 
 class CSV0DModelParser(object):
     '''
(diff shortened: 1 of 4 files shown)


FinbarArgus:
Follow-up filed as #398 for wiring `rtol`/`atol` through to the generated C++, so the two settings this PR removes from the cpp schema come back the moment the generator reads them.

PR #367, opened 2026-08-07
Developing pyMC for UQ

MOH-Shafizadegan:
Features #195, #193, #179.

Several tests for UQ functionality is included and all are passed in the current branch.

Issue #399, opened 2026-08-12
cpp generation fails with KeyError: 'dt_solver' — the cpp solver_info defaults never seed it


FinbarArgus:
Found while working on #398; not caused by it and not fixed there.

## Symptom

`tests/test_autogeneration.py::test_generate_cpp_model_succeeds` fails on current master:

```
KeyError: 'dt_solver'
```

It is marked `slow`, so it is deselected from the default `-m "not slow"` run and the breakage is not visible there.

## Cause

The cpp branch of `script_generate_with_new_architecture` requires the key with no fallback:

```python
dtSolver = solver_info['dt_solver']          # src/scripts/script_generate_with_new_architecture.py
```

but `get_solver_info_default('cpp')` never seeds it:

```python
if model_type == 'cpp':
    return {
        'solver': 'CVODE',
        'MaximumStep': 0.001,
        'MaximumNumberOfSteps': 5000,
        'rtol': 1e-8,
        'atol': 1e-8,
    }
```

`dt_solver` is one of `_FRAMEWORK_SOLVER_INFO_KEYS`, so it is *permitted* for every solver — but permitted is not the same as defaulted. Any cpp config that does not set it explicitly in `user_inputs.yaml` fails at generation time with a bare `KeyError`, which does not say which setting is missing or where to put it.

## Suggested fix

Either seed `dt_solver` in the cpp defaults (`0.001` matches the `MaximumStep` default it effectively replaces for this backend), or read it with an explicit default and a clear error naming the setting when it is genuinely required. The first is probably right: the cpp path always needs a solver step, so there is no configuration in which its absence is meaningful.

Whichever way, the test above should pass again, and it is worth checking whether it is the only cpp coverage that would have caught this — it is currently `slow`, so a regression here can sit unnoticed for a while.

## Note on scope

#398 wires `rtol`/`atol` through to the generated C++ and touches the same few lines of this function, but deliberately leaves this alone: it is a separate defaulting bug, and fixing it inside that PR would have mixed an unrelated behaviour change into a settings-plumbing change.

PR #403, opened 2026-08-12
feat(uq): KDE and Poisson likelihood cost functions (#367 step 2)

FinbarArgus:
Second of the stack bringing in #367 (pyMC for UQ). Original work by @MOH-Shafizadegan.

> Stacked on #402 — its commit is included here and the diff shrinks to the last commit once #402 merges.

## What

Two likelihood costs, both `@is_MLE`:

- **`kernel_density_estimation`** — scores the model output against a `gaussian_kde` of the observed samples. For a target known only as a set of measurements rather than as a named distribution. Unlike `multimodal_gaussian` it needs no assumption about how many modes there are or where they sit, which is what makes the bimodal recovery test later in the stack possible.
- **`poisson_MLE`** — the Poisson negative log-likelihood for count data.

## A sign bug fixed on the way in

Both are `@is_MLE`, so paramID reads them as `ln L = -cost` and sums them alongside `gaussian_MLE`. The sign is therefore part of the contract — and #367 had `poisson_MLE` returning

```python
cost = (k * np.log(lam) - lam) * weight     # this is the LOG-likelihood
```

which is *maximised* at `lambda == k`. Minimising it would have driven the rate **away** from the observed count in both directions, and nothing would have failed loudly — the calibration would just have converged somewhere wrong.

Corrected to `lambda - k*log(lambda)`, and pinned by a test that scans the cost curve over `lambda` and asserts the minimum sits at `k`. Reverting to #367's formula fails three of the new tests, including that one.

## Also hardened relative to #367

- The Poisson rate is clipped off zero. An unclipped `log(0)` returns `-inf`, which poisons the summed cost for *every other* observable and turns one bad parameter set into an uninformative total.
- The KDE result is coerced to a plain `float` — `logpdf` returns an array even for a single point, so the weighted sum over observables would otherwise stop being 0-d.
- The validated `data_points` array is the one handed to `gaussian_kde` (#367 revalidated it, then passed the raw list).
- Bad `prob_dist_params` raise a message naming the missing key, instead of an `IndexError` further in.

## Tests

New `tests/test_uq_cost_funcs.py`, 11 fast unit tests: both costs are discoverable as `is_MLE` and not differentiable; the Poisson NLL matches the analytic value and is minimised at `k`; the KDE cost matches `scipy.stats.gaussian_kde` directly and scores both modes of a bimodal sample set better than the gap between them; bandwidth and weight are honoured; unusable inputs raise; and all three MLE costs agree on which direction is "better" — they get summed into one cost, so a disagreement would cancel them out.

`-m unit`: 299 passed, 1 skipped.

### DIFF
--- a/funcs_user/cost_funcs_user.py
+++ b/funcs_user/cost_funcs_user.py
@@ -99,6 +99,72 @@ def multimodal_gaussian(output, prob_dist_params, weight):
     return cost
 
 
+@is_MLE
+def kernel_density_estimation(output, prob_dist_params, weight):
+    """Negative log-likelihood under a kernel density estimate of the observed samples.
+
+    For a target that is known only as a set of measurements rather than as a named
+    distribution: the samples are turned into a smooth density with
+    ``scipy.stats.gaussian_kde`` and the model output is scored against it. Unlike
+    ``multimodal_gaussian`` this needs no assumption about how many modes there are, or
+    where they sit.
+
+    prob_dist_params: ``{"data_points": [...], "bandwidth": <optional>}``. ``bandwidth`` is
+    passed straight to gaussian_kde's ``bw_method`` ('scott', 'silverman', a scalar, or a
+    callable), defaulting to Scott's rule.
+    """
+    if hasattr(output, "__len__") and np.size(output) > 1:
+        raise ValueError(
+            "kernel_density_estimation cost function is not implemented for series data")
+
+    if not isinstance(prob_dist_params, dict) or "data_points" not in prob_dist_params:
+        raise ValueError(
+            "prob_dist_params for kernel_density_estimation in obs_data.json must be a dict "
+            "with a 'data_points' entry (and optionally 'bandwidth')")
+
+    data_points = np.asarray(prob_dist_params["data_points"], dtype=float)
+    if data_points.size == 0:
+        raise ValueError(
+            "data_points for kernel_density_estimation in obs_data.json cannot be empty")
+
+    from scipy.stats import gaussian_kde
+
+    kde = gaussian_kde(data_points, bw_method=prob_dist_params.get("bandwidth", "scott"))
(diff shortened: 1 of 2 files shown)


FinbarArgus:
**The param_id CI failure was flaky, not caused by this PR.** I re-ran the identical commit (`5efdd88`, no changes) and `test-param-id-mpi` **passed in 23m38s**; `test-param-id-serial` passes too.

The original failure was:

```
RuntimeError: Myokit solver requested but Myokit is not available
```

on a runner whose own pip log a minute earlier reads `Successfully installed ... myokit-1.39.2`. The job died at 1m45s, at 2% of the tests. This PR adds two pure functions to `funcs_user/cost_funcs_user.py` and one new test file that the MPI job does not even collect (it runs only `test_param_id.py` and `test_sensitivity_analysis.py`), so there was no path from the change to Myokit availability.

Most likely cause of the transient: myokit writes `~/.config/myokit/myokit.ini` on first import, and the job runs `mpiexec -n 2`, so two ranks race to create it on a fresh runner.

**What genuinely was wrong is that the message pointed away from the problem.** `solver_wrappers/__init__.py` imported the myokit backend under a bare `except:`, so any failure — transient or otherwise — became "Myokit is not available", i.e. indistinguishable from it not being installed. Fixed separately in #410, which records the reason and includes it in the error; that would have answered this in one look instead of a log hunt.

### Remaining red checks are pre-existing on master

`test (3.9 / 3.10 / 3.11)` also fail on master at the same commit this branch is based on (`3e8e233`). Their failures here are:

- `test_generate_cpp_model_succeeds` — `KeyError: 'dt_solver'`, which is #399
- `test_generate_python_model_succeeds[SN_simple]` and `test_python_BDF_solver[SN_simple]` — the libCellML Analyser limitation, #151
- `test_aadc_semi_implicit_signed_forward_tracks_cvode` — AADC

None of them touch cost functions. This PR is ready as-is.

Issue #344, opened 2026-08-03
CMA-ES shows superlinear speedup in the 3compartment scaling benchmark, so the runs are not doing equal work


FinbarArgus:
## What

In the `3compartment` core-scaling sweep, CMA-ES reports a speedup above the physical ceiling. Three independent sweeps on an otherwise idle machine:

| sweep | 1 core | 2 cores | 4 cores | 8 cores | speedup 1→8 |
|---|---|---|---|---|---|
| A | 213 s | 62.7 s | 30.8 s | 9.4 s | **22.76×** |
| B | 116 s | 94 s | 9 s | 7.0 s | **16.61×** |
| C | 179 s | 68 s | 22 s | 12 s | **15.01×** |

8 cores cannot give more than an 8× speedup, so this is not parallel scaling. It reproduces in all three sweeps, so it is not noise either — it is a property of the benchmark.

## Why it matters

The scaling table is published in the docs and used for the CUFLynx paper. A superlinear row invites the reading that CUFLynx parallelises better than linearly, which is not what is happening.

## Likely cause

The runs are almost certainly not doing **equal work** at each core count. `three_compartment_config` sets `no_new_starts_on_convergence: False` specifically so every start runs at every core count — but that governs multi-start, not CMA-ES. If CMA-ES terminates on a convergence or patience criterion, the number of cost evaluations it performs can differ with rank count, and the wall-clock ratio then measures "how soon it happened to converge" rather than throughput.

The shape of sweep B supports this: 1→2 cores is 116 s → 94 s (barely any speedup) and then 2→4 is 94 s → 9 s. That is not a parallelisation curve.

Its absolute times are also small (7-12 s at 8 cores against 116-213 s at 1 core), so a run that stops a few generations early dominates the ratio.

## Suggested investigation

- Log the actual number of cost-function evaluations CMA-ES performs at each core count. If they differ, the speedup is not measuring scaling.
- Check which termination criteria CMA-ES honours in `optimisers.py` and whether any is reachable inside the configured budget for this model.
- If it is convergence-driven, either fix the evaluation count for the scaling benchmark (the CMA-ES analogue of `no_new_starts_on_convergence: False`) or exclude CMA-ES from the scaling table with a note.

## Context: the other optimisers are stable in shape but not in value

Same three sweeps:

| optimiser | speedup 1→8 (A / B / C) |
|---|---|
| `genetic_algorithm` | 3.63× / 5.53× / 3.94× |
[...]

Issue #193, opened 2026-04-21
Fixing cost function and log likelihood for MCMC


MOH-Shafizadegan:
Unlike parameter identification, for MCMC, all features forming the likelihood should have same weight to represent the true probability distribution.

This should be handled and a warning flag should be sent.

FinbarArgus:
Make sure we only set non-zero weights to 1. zero weights should still be ignored, if for example it is only there for validation.

MOH-Shafizadegan:
Additionally, there is no need for multiplication with 0.5 inside `get_lnlikelihood_from_params()` ([here](https://github.com/physiomelinks/circulatory_autogen/blob/master/src/param_id/paramID.py#L1637)) for generic multimodal distribution. The multplication by 0.5 should be handled inside the cost function itself for Gaussian distributions.

FinbarArgus:
@MOH-Shafizadegan can you please prioritise checking this and confirming it can be closed with ffef14? Or letting me know if you have a branch that adds to the fix.

PR #404, opened 2026-08-12
feat(uq): let the GA minimise a log posterior, safely (#367 step 3)

FinbarArgus:
Third of the stack bringing in #367 (pyMC for UQ). Original work by @MOH-Shafizadegan.

> Stacked on #402 and #403 — their commits are included here and the diff shrinks to the last commit once those merge.

## What

`objective_function` (`'cost'` | `'likelihood'`) lets the genetic algorithm search the same objective MCMC samples — which is how a GA is used to find a starting point for UQ. Only the GA reads it: it is the one optimiser that needs no gradient of that objective.

Getting there needed two fixes.

## 1. The likelihood objective has to be negated

`get_lnlikelihood_lnprior_from_params` returns a **log posterior** — higher is better — and every optimiser here **minimises**. #367 handed it over unnegated:

```python
self.objective_method = lambda params: self.param_id_obj.get_lnlikelihood_lnprior_from_params(params)
```

That would have made the GA search for the *least* probable parameters. And worse: an out-of-prior point returns `-inf`, which a minimiser scores as **the best point there is**, so the population would have been driven straight out of the prior support. Negated, that same `-inf` becomes `+inf` — exactly the "reject and resample" signal the population loop already handles.

## 2. The survival weighting — and why I did **not** take #367's version

`cost**-1` assumes a positive cost. A negative log posterior goes negative wherever the posterior density exceeds 1, and a negative weight is not a probability. Even where it stays positive it barely discriminates: `1/800` vs `1/830` is almost uniform selection, i.e. a GA that has stopped selecting.

#367 replaced the rule with `exp(-cost)` **for every run**. Measured on a real 3compartment GA population (costs 5.32 → 17.34, taken from the existing fast GA test), using perplexity as the effective number of distinct survivors:

| rule | effective survivors (of 6) |
|---|---|
| current `cost**-1` | **5.41** |
| #367 `exp(-cost)` | **1.27** |

That is near-deterministic selection — the population diversity a GA exists to keep, gone, on every ordinary calibration. It also interacts with the GA scaling benchmark (#344).

So this PR **keeps the inverse-cost rule wherever it is valid** — every ordinary calibration, where a weighted cost function returns a strictly positive cost — and uses Boltzmann selection only as the fallback for objectives it cannot handle. The fallback is shifted to start at zero (`exp(-cost)` underflows to zero for every member past about 745, and softmax is shift-invariant so this changes no probability) and scaled by the population spread, so selection pressure does not depend on the objective's absolute magnitude.

**Existing calibrations are bit-for-bit unaffected.** The 3compartment GA test returns the identical best cost, `5.3206291895193765`, before and after.

## 3. Relative convergence tolerance

`use_relative_tolerance` / `relative_tolerance` add a fractional stall test alongside the absolute `cost_convergence` one, which a large-magnitude objective never trips (a log posterior in the hundreds never moves by less than 1e-4, so the run only ever stops on the generation budget). Either criterion counts, so enabling it can only make a run stop sooner, never later.

All three options are registered in `PARAM_ID_METHODS['genetic_algorithm']`, so CUFLynx offers them and `test_param_id_method_options_match_optimiser_reads` covers them.

## Tests

New `tests/test_ga_objective_and_selection.py`, 14 fast unit tests: the default objective is the cost function; the likelihood objective is negated so lower is still better; an out-of-prior point becomes `+inf` rather than the best point; an unknown objective is rejected; an ordinary positive cost keeps the exact inverse-cost weighting; selection does not collapse onto a single member (with #367's form asserted to be the collapse this guards against); the fallback survives log-posterior magnitudes; a non-finite member neither wins nor poisons the others; and the relative tolerance catches a stall the absolute one misses without dividing by zero.

Verified `not slow and not compare_optimisers`: **625 passed, 1 failed** — `test_python_BDF_solver[SN_simple]`, the known libCellML Analyser limitation (#151), which fails identically on master.

## Next in the stack

#193 equal feature weighting for MCMC → diagnostics → pyMC backend.

### DIFF
--- a/src/param_id/optimisers.py
+++ b/src/param_id/optimisers.py
@@ -126,6 +126,92 @@ class GeneticAlgorithmOptimiser(Optimiser):
     _POPULATION_KEYS = ('num_elite', 'num_survivors', 'num_mutations_per_survivor',
                         'num_cross_breed')
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.objective_function = self.optimiser_options.get('objective_function', 'cost')
+        if self.objective_function not in ('cost', 'likelihood'):
+            raise ValueError(
+                f"Invalid objective_function: {self.objective_function!r}. "
+                f"Must be 'cost' or 'likelihood'.")
+
+    def _objective(self, param_vals):
+        """The scalar this GA minimises, for the selected ``objective_function``.
+
+        ``'likelihood'`` is **negated**: get_lnlikelihood_lnprior_from_params returns a log
+        posterior, where higher is better, and every optimiser here minimises. Handing it over
+        unnegated (as #367 did) would have made the GA search for the *least* probable
+        parameters -- and worse, an out-of-prior point returns -inf, which minimisation would
+        score as the best point there is. Negated, that same -inf becomes +inf, which is exactly
+        the "reject and resample" signal the population loop below already handles.
+        """
+        if self.objective_function == 'likelihood':
+            return -self.param_id_obj.get_lnlikelihood_lnprior_from_params(param_vals)
+        return self.param_id_obj.get_cost_from_params(param_vals)
+
+    def _loss_has_stalled(self, cost, last_loss):
+        """Whether this generation counts towards ``max_patience``.
+
+        The absolute ``cost_convergence`` test alone is scale dependent: a likelihood objective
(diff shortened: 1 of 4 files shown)


FinbarArgus:
@MOH-Shafizadegan please review

Issue #349, opened 2026-08-05
cost_calc indexes per-type weight vectors by compacted index, but they are built per data_item row


FinbarArgus:
`cost_calc` reads each per-type weight vector with the **compacted, per-type index**, but `process_protocol_and_weights` builds those vectors with **one entry per `data_items` row**. The two only agree when the items of a given type occupy the leading rows, so an obs_data that interleaves types silently mis-scores.

### Where

`src/parsers/PrimitiveParsers.py:2570` builds a full-length vector, zero where the row is not that type:

```python
full_weights = np.where(mask & (df["data_type"] == data_type), df["weight"], 0.0)
weight_map[exp_idx][this_sub_idx] = full_weights
```

`src/param_id/paramID.py:1951` then indexes it with `const_idx`, the compacted counter:

```python
for const_idx in range(len(const)):
    obs_idx = self.obs_info['const_idx_to_obs_idx'][const_idx]
    if updated_weight_const_vec[const_idx] != 0:            # <- row index expected here
        cost += cost_funcs_dict[self.cost_type[obs_idx]](...)  # <- but obs_idx used one line down
```

The mapping needed is already in hand and already used on the very next line for `cost_type` — it just isn't used for the weight lookup. The same pattern applies to the `series`, `amp`, `phase` and `prob_dist` blocks.

### Reproducer

Two `series` items followed by one `constant` of weight 7.0:

```python
from parsers.PrimitiveParsers import ObsAndParamDataParser
items = [
    {"variable":"a/x","data_type":"series","unit":"mV","operands":["a/x"],
     "value":[1.0,2.0],"std":1.0,"weight":1.0,"obs_dt":0.01,"name_for_plotting":"s1"},
    {"variable":"a/y","data_type":"series","unit":"mV","operands":["a/y"],
     "value":[1.0,2.0],"std":1.0,"weight":1.0,"obs_dt":0.01,"name_for_plotting":"s2"},
    {"variable":"a/z","data_type":"constant","unit":"mV","operands":["a/z"],
     "value":5.0,"std":1.0,"weight":7.0,"name_for_plotting":"c1"},
]
obs = {"data_items": items, "protocol_info": {"pre_times":[0.0],"sim_times":[[1.0]]}}
p = ObsAndParamDataParser()
parsed = p.parse_obs_data_json(obs_data_dict=obs, pre_time=0.0, sim_time=1.0)
[...]

Issue #315, opened 2026-07-26
casadi_python + gradient-free calibration: numpy operands hit the casadi operation-func backend ('numel')


FinbarArgus:
## Summary

Running a **gradient-free** method (genetic algorithm) with **`model_type =
casadi_python`** crashes evaluating any observable that uses a backend-dispatched
reduction such as `mean`:

```
AttributeError: 'numpy.ndarray' object has no attribute 'numel'
  ...
  paramID.py:2077  obs = self.operation_funcs_dict[...](*operands_outputs[JJ], **kwargs)
  operation_funcs.py:43  return mb.mean(x)
  math_backend.py:150   return ca.sum(x) / x.numel()   # x is a numpy.ndarray
```

Repro: `casadi_python` + `genetic_algorithm`, obs_data with a data_item whose
`operation` is `mean` (e.g. the 3compartment model).

## Root cause

For `casadi_python`, the operation/cost func dicts are built **once, in casadi
mode** (`paramID.py:1149-1155`):

```python
if self.model_type == "casadi_python":
    mode = "casadi"
...
self.operation_funcs_dict = self.sfp.get_operation_funcs_dict(mode)
```

But the **numeric** cost path — `get_cost_from_params` → … →
`get_obs_output_dict(operands_outputs, is_symbolic=False)` (`optimisers.py:248`
for GA) — applies that **casadi-backend** dict to **numpy** operands
(`paramID.py:2077`). `CasadiBackend.mean` is `ca.sum(x) / x.numel()`
(`math_backend.py:150`), and numpy arrays have no `.numel()`, so it raises.

The `is_symbolic=False` branch has no numpy-mode operation dict to fall back to —
there's only the single casadi one. This surfaced after the operation/cost funcs
were switched to the backend `mb` dispatch (#199); a previous numpy-native `mean`
tolerated numpy operands.

## Impact

Any gradient-free method (genetic_algorithm, CMA-ES, Bayesian) on `casadi_python`,
for observables using a backend-specific reduction (`mean` is the clear one;
`max`/`min`/`max_minus_min` go through `ca.mmax`/`ca.mmin` which may also reject
numpy). The symbolic/AD path (`sp_minimize` with do_ad) is unaffected because its
operands are casadi symbols.

## Suggested fix

Use **numpy-mode** operation/cost funcs for the numeric (`is_symbolic=False`)
evaluation path even when `model_type == casadi_python` — e.g. build both a numpy
[...]

PR #405, opened 2026-08-12
fix(uq): weight every likelihood feature equally for MCMC (#193, #367 step 4)

### LINKED ISSUE #193: Fixing cost function and log likelihood for MCMC

FinbarArgus:
Closes #193. Fourth of the stack bringing in #367 (pyMC for UQ). Original work by @MOH-Shafizadegan.

> Stacked on #402, #403, #404 — their commits are included here and the diff shrinks to the last commit once those merge.

## Why

Calibration weights are a modelling choice: they say which features the optimiser should care about most. **A posterior is not a modelling choice.**

Under `ln L = -cost`, a weight `w` on a feature raises its likelihood term to the power `w` — which is the same claim as `w` independent observations of it. A feature weighted 10 shrinks the posterior as if it had been measured ten times, so the credible intervals that come out are **not the ones the data supports**. The relative weighting between features distorts their trade-off the same way.

## What

`OpencorMCMC` flattens every non-zero weight to 1.

**Zero weights are preserved.** A zero does not mean "unimportant" — it means the observable is not part of that sub-experiment. Reinstating it would add a feature the user excluded, and would also desynchronise the cached `_num_weighted_obs_by_exp_sub` denominator, which counts non-zeros. The non-zero count is unchanged by design, and a test pins that.

## How — one seam, not two copies of the cost

The five weight-vector reads are extracted out of `cost_calc` into `_cost_weight_vectors(exp_idx, sub_idx)`, and `OpencorMCMC` overrides **only that**.

#367 instead duplicated the entire body of `cost_calc` into `OpencorMCMC`. Two copies of the cost machinery is exactly how a fix lands on calibration and silently misses UQ — the cost path has changed repeatedly since #367 forked (#315 symbolic/numeric dispatch, #349 the weight-indexing fix). So there is a test asserting `OpencorMCMC.cost_calc is OpencorParamID.cost_calc`, i.e. that the duplication has not come back.

## Warning

Warns **once**, not once per cost evaluation, and only when the flattening actually changed something — so a user who tuned weights for a calibration and then ran UQ on the same obs_data finds out they no longer apply, without drowning the run in output.

```
WARNING: obs_data weights are ignored for UQ -- every feature entering the likelihood is
weighted 1. A weighted likelihood is not a posterior: a weight w on a feature is the same
claim as w independent observations of it, so it would shrink the credible intervals by a
factor the data does not support (issue #193). Weights of 0 still exclude an observable.
```

## Tests

New `tests/test_uq_equal_weights.py`, 6 fast unit tests: calibration keeps the user's weights untouched; UQ flattens every non-zero to 1; zeros survive and the non-zero count is unchanged; the warning fires once and not per evaluation; it stays silent when weights were already uniform; and `OpencorMCMC` inherits `cost_calc` rather than overriding it.

Verified `not slow and not compare_optimisers`: **631 passed, 1 failed** — `test_python_BDF_solver[SN_simple]`, the known libCellML Analyser limitation (#151), which fails identically on master.

The slow posterior-recovery test from #367 that exercises this end to end lands with the pyMC backend in the final step, once its obs_data fixtures come across.

## Next in the stack

MCMC diagnostics (rhat / ESS / autocorrelation) → pyMC backend.

### DIFF
--- a/src/param_id/paramID.py
+++ b/src/param_id/paramID.py
@@ -2216,6 +2216,23 @@ class OpencorParamID():
             return None
         return raw[obs_idx] or None
 
+    def _cost_weight_vectors(self, exp_idx, sub_idx):
+        """The five per-data_item weight vectors this sub-experiment's cost is built from.
+
+        A single seam so a subclass can change what weighting the cost uses without
+        reimplementing cost_calc -- OpencorMCMC flattens them, because a weighted likelihood is
+        not a posterior (issue #193).
+
+        Returns them in the order (const, series, amp, phase, prob_dist).
+        """
+        return (
+            self.protocol_info["scaled_weight_const_from_exp_sub"][exp_idx][sub_idx],
+            self.protocol_info["scaled_weight_series_from_exp_sub"][exp_idx][sub_idx],
+            self.protocol_info["scaled_weight_amp_from_exp_sub"][exp_idx][sub_idx],
+            self.protocol_info["scaled_weight_phase_from_exp_sub"][exp_idx][sub_idx],
+            self.protocol_info["scaled_weight_prob_dist_from_exp_sub"][exp_idx][sub_idx],
+        )
+
     def cost_calc(self, obs_dict, exp_idx=0, sub_idx=0, is_symbolic=False):
 
         # Symbolic cost terms use the casadi-mode cost funcs; numeric ones the numpy-mode funcs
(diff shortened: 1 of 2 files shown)


FinbarArgus:
@MOH-Shafizadegan please review

FinbarArgus:
**Found and fixed — this one was a real bug in the PR, not a flake.**

`test-param-id-serial` failed on `test_param_id_3compartment_extra_ops_succeeds` with `KeyError: 'volume_sum_module.t'`. That key error is a red herring; the actual defect is one line above it in the log:

```
WARNING: best cost 0.03774134719209266 is not close to cost check 0.10579019655881715
```

## What went wrong

Flattening the weights for the likelihood also flattened **every other cost the MCMC engine computes** — including the one `OpencorMCMC.run()` writes to `best_cost.npy`.

`best_cost` is a *calibration* artifact: `plot_param_id` builds a plain calibration engine, `simulate_once` re-derives the cost with the real weights and checks it against the saved value, and the calibration itself optimised the weighted cost. Written on the flat scale it is a different quantity under the same name, and the two disagree by exactly whatever the weights were.

Reproduced locally on 3compartment — **0.0381 saved vs 0.1063 recomputed**, against CI's 0.0377/0.1058:

| | best_cost saved | cost recomputed |
|---|---|---|
| before | 0.0381 | 0.1063 |
| after | 0.0937 | 0.0937 |

**A worse consequence than the mismatch:** `mcmc_best_cost` was compared against `original_best_cost` to decide whether to *overwrite* `best_param_vals`. On the flat scale that comparison could replace a genuinely better calibration fit using an objective the calibration never optimised. Both sides are now weighted like the calibration.

The flattening still applies to the likelihood being sampled — that is the whole point of #193. `calibration_weighting()` only suspends it for costs that leave the sampler.

## Why local runs were green

The crash sits inside a debug block that only executes when the cost check mismatches **and** `all_outputs_with_best_param_vals_exp_0.npz` exists. That file is present in CI and absent in my local runs, so locally the same bug degraded to a warning and the test passed. Worth knowing for the next one of these.

That block also had its own latent bug: it indexed the *live* run's outputs with every key from the *saved* npz and raised when they differed. A diagnostic that raises destroys the information it exists to print, and it runs precisely when something is already wrong — missing keys are now reported and skipped.

## Verification

- The failing test now reports `cost should be 0.09371146570564633` / `cost check after single simulation is 0.09371146570564633` — identical.
- Reverting only the fix reproduces the mismatch, so the regression and the fix are both pinned.
- This also silences a mismatch I had seen but not recognised on `test_param_id_3compartment_succeeds` (`0.454` vs `2.209`); it now matches at `2.209`.
- 4 new tests in `tests/test_uq_equal_weights.py`, including one that pins **both** call sites in `run()` rather than just the helper.
- Full `not slow and not compare_optimisers`: **635 passed, 1 failed** — the known SN_simple libCellML limitation (#151).

Issue #341, opened 2026-08-03
Deprecate built-in result plotting in favour of a generated, user-editable plot_outputs script


FinbarArgus:
Related: **physiomelinks/CUFLynx#158** (the CUFLynx side, implemented there).

## What has changed

Automatic plotting of results is now done through **CUFLynx**, and the creation of savable plots is migrating to the **CUFLynx plotting script** — a standalone `plot_outputs.py` written next to a run's outputs. The point is that the user can **adapt the plots to their own needs**: it is ordinary matplotlib in a file they own, so changing a colour, a layout, an axis or adding a panel is an edit, not a fork of CA.

The direction is that **CA's own plot generation is deprecated in favour of automatically generating that user-editable `plot_outputs` script**.

## What CUFLynx's script now draws, and from what

Everything comes from files a run already leaves behind. No model, no solver, no CA import — so it also works on a run directory copied to another machine:

| figure | read from |
|---|---|
| best fit vs. observations | `all_outputs_with_best_param_vals_exp_*.npz` + `obs_data.json` |
| calibration error bars (percent and std) | `percent_error_vec.npy`, `std_error_vec.npy` + `obs_data.json` |
| cost / parameters vs generation | `best_cost_history.csv`, `best_param_vals_history.csv` |
| sensitivity, UQ | `results.json` |
| simulation traces | `simulation.json` |

This is the part worth noting for CA: **`param_id/plot_outputs.py` cannot do this**, because `_fetch_best_fit_data()` re-runs the model through a live `paramID` object to recover the traces. The figures it produces are reachable from the saved artefacts alone — which is what makes generating a script viable at all, and what makes a saved run re-plottable later without re-simulating it.

## Suggested direction for CA

1. **Keep writing the artefacts.** `all_outputs_with_best_param_vals_exp_*.npz`, `percent_error_vec.npy`, `std_error_vec.npy`, the histories and the dated `obs_data.json` copy are now an interface, not an implementation detail. Their names and shapes are what the generated script binds to.
2. **Emit the script instead of the figures.** At the end of a param-id run, write a `plot_outputs.py` beside the outputs (as CUFLynx does) rather than 15 files of `.png`/`.pdf`/`.eps`. Running it reproduces them; editing it changes them.
[...]

Issue #167, opened 2026-03-24
sensitivity analysis naming convention double underscores not supported


Tasjabeulens:
````
"name_for_plotting": "u_{A_{R}}"
FileNotFoundError: [Errno 2] No such file or directory: \sensitivity\\phlynx-export-3C-BVC\\u\\_{A\\_{R}} - experiment0, subexperiment0_n2_First_order_idx.png'
````

MOH-Shafizadegan:
This isn't an issue on Linux! Seems to be a problem only on Windows.

FinbarArgus:
Fixed in d1279e2 (on `master`). Sensitivity-analysis plot filenames are now routed through `sanitize_for_filename()` in `src/sensitivity_analysis/sobolSA.py`, so LaTeX-ish `name_for_plotting` values such as `u_{A_{R}}` (and spaces/commas) can no longer produce a path that fails to write on Windows. Regression test in `tests/test_issue_fixes.py`.

PR #406, opened 2026-08-12
feat(uq): MCMC convergence diagnostics -- split R-hat, ESS, chain plots (#367 step 5)

FinbarArgus:
Fifth of the stack bringing in #367 (pyMC for UQ). Original work by @MOH-Shafizadegan.

> Stacked on #402–#405 — their commits are included here and the diff shrinks to the last commit once those merge.

## Why

An MCMC run reported a posterior with nothing to say whether the chain had converged, so a posterior from walkers that never mixed looked exactly like one from walkers that did.

## What

Per-parameter **split R-hat** and **effective sample size**, a summary table (mean / sd / 3% / 97% / ess / r_hat), an **autocorrelation** plot and a running **chain-average** plot. `plot_mcmc` now prints the table and states the verdict against the 1.01 R-hat threshold rather than leaving the reader to apply it — a diagnostic nobody reads is not a diagnostic.

```
parameter                           mean          sd          3%         97%       ess    r_hat
...
WARNING: r_hat > 1.01 for ['q_{sbv}', 'C_{ao}', 'E_{lvA}', 'E_{lvB}'] -- the walkers have not
mixed. Run more steps before trusting the posterior.
```

(That is the real output of the existing 3compartment MCMC test, whose DEBUG chain is 5 steps long — so the diagnostic is doing its job on the first run it saw.)

## No new dependency

#367 imported `arviz`, `seaborn` **and** `pytensor` at module level in `paramID.py` for these. `paramID.py` is imported by every calibration run, and none of those three are CA dependencies — so `paramID.py` would not have imported at all for most users, and the diagnostics would have been unavailable in exactly the environments that need them. (`arviz` is not installed in the OpenCOR pythonshell this repo runs under.)

R-hat and ESS are short, standard formulas, and `emcee` — already a dependency — supplies the autocorrelation. So they are computed directly from numpy and emcee, and the tests below actually run rather than skipping.

## Two things beyond a port

- **R-hat is the split form.** `az.summary` is split too, but the distinction is worth being explicit about: a walker drifting steadily has a large *within*-chain variance, which is exactly what makes an unsplit ratio look fine. There is a test that builds precisely that chain and asserts it is caught.
- **One summary, computed once.** #367 rebuilt the same arviz dataset and re-ran `az.summary` inside each of `get_posterior_stats`, `calc_effective_sample_size` and `calc_rhat`, so the expensive part ran three times to fill one table.

ESS is also capped at the draw count — a chain cannot carry more independent information than it has draws, whatever the autocorrelation estimator does on a short trace.

## Tests

New `tests/test_uq_diagnostics.py`, 15 fast unit tests against chains whose answer is known by construction: R-hat ≈ 1 for a well-mixed chain, > 1.5 for walkers parked in different places, > 1.5 for a steadily drifting chain, and `nan` rather than a guess on a chain too short to split; ESS ≈ the draw count for independent draws and < 10% of it for a strongly autocorrelated one, never exceeding the draw count; the summary recovers a known mean and sd; the verdict is printed both ways; labels fall back to indices when names do not match; both plots are written, the autocorrelation plot reports whether it decayed inside ±0.1, the chain-average plot skips a chain shorter than its window, and a single-parameter chain still plots (`plt.subplots` returns a bare Axes for one row, which is a routine way for per-parameter plotting to break).

Verified `not slow and not compare_optimisers`: **646 passed, 1 failed** — `test_python_BDF_solver[SN_simple]`, the known libCellML Analyser limitation (#151), which fails identically on master.

## Deliberately not included

#367's three seaborn posterior-predictive plots — `plot_boxplots_for_predictions`, `plot_distribution_grid`, `get_prior_pdf` — are **not** here. They are result presentation rather than convergence diagnostics, they need the prediction machinery to re-simulate from posterior draws, and #341 is an open decision to deprecate built-in result plotting in favour of a generated, user-editable script. Adding three more built-in plots cuts against that, so they are worth landing (or not) as their own change once #341 is settled.

## Next in the stack

The pyMC backend, and the slow posterior-recovery tests.

### DIFF
--- a/src/param_id/paramID.py
+++ b/src/param_id/paramID.py
@@ -951,6 +951,14 @@ class CVS0DParamID():
         # Also check autocorrelation times for mcmc chain
         tau = self.calculate_autocorrelation_time(samples)
 
+        # Per-parameter posterior summary with ESS and split-R-hat, and the two chain-diagnostic
+        # plots. Printed rather than only returned: R-hat and ESS are the numbers that say
+        # whether the posterior above is trustworthy at all, so a run should not be able to
+        # finish without stating them.
+        self.print_convergence_diagnostics(samples)
+        self.plot_autocorrelation(samples)
+        self.plot_chain_avg(samples)
+
         # check geweke convergence
         if not self.DEBUG:
             # the chain is too short when running debug to do geweke diagnostics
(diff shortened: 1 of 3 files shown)


FinbarArgus:
Scope update: the three seaborn posterior-predictive plots **are now included** (commit `ad5a263`), so the "Deliberately not included" section in the description above no longer applies.

`plot_boxplots_for_predictions`, `plot_distribution_grid` and `get_prior_pdf` are ported, with seaborn imported inside each rather than at module level. Four fixes on the way in:

- **`get_prior_pdf` is derived from the engine's own `get_lnprior_from_params`** rather than reimplemented. The priors are independent per parameter, so evaluating the log prior along one axis with the others held at their best fit recovers that parameter's prior up to a constant, which the normalisation removes. #367 restated the *old* hardcoded defaults (`lambda=1`, `sigma=(max-min)/6`, `mu=(max+min)/2`) — so a plotted prior would have silently disagreed with the one actually being sampled for anyone using `prior_mean` / `prior_std` / `prior_origin` / `prior_scale`, or the `unbounded` flag.
- **`plot_distribution_grid` is called once**, after the per-feature loop. #367 called it *inside*, redrawing the whole grid once per feature and keeping only the last.
- **Figure names go through `sanitize_for_filename`.** A `name_for_plotting` is LaTeX-ish (`u_{A_{R}}`), and `feature.replace(' ', '_')` leaves the braces and slashes that do not survive as a filename — the same defect #167 fixed for the sensitivity plots.
- **`n_sims` larger than the sample count no longer raises** — `np.random.choice(replace=False)` cannot draw more than it has.

The predictive values are also written to `posterior_predictions.csv` in long format, so they can be re-plotted without paying for the simulations again.

12 new tests in `tests/test_uq_posterior_plots.py`. Full `not slow` run: **658 passed, 1 failed** — the known SN_simple libCellML limitation (#151).

Issue #179, opened 2026-04-14
Create quantitative tests for param ID and MCMC where we have synthetic ground truths and probability distributions.


FinbarArgus:

FinbarArgus:
@MOH-Shafizadegan have you done this already?

MOH-Shafizadegan:
Yes, there is a simple benchmark model with analytical solution for parameter posteriors. This could be used to compare both Laplace approximation and MCMC results to the analytical solution. 

All I need to do is to format the code into a pytest.

FinbarArgus:
Great! Could you add testing for estimating the MAP in this benchmark problem with each of our methods too, genetic algorithm, CMAES, gradient decent?

Then every time we have a new method we can add a test for it with this benchmark problem.

FinbarArgus:
These tests exist now, across all three of the things this asked for.

**Posterior recovery against a synthetic ground truth** — `tests/test_UQ.py` opens with "does the posterior recover a known answer? (#179)" and asserts the sampled posterior against `ANALYTICAL_SOLUTION` / `ANALYTICAL_STD` (`:48-49`, via `_assert_posterior_recovers_the_truth` at `:114-121`). The CI-speed equivalent is `tests/test_uq_on_emulator.py:230-294`, which does the same for **both** emcee and pymc.

**Probability-distribution ground truths** — `tests/test_distribution_ground_truth.py`.

**Per-optimiser MAP recovery** — `benchmarks/benchmark_specs.py:194-226`: `assert_fitzhugh_nagumo` asserts `FHN_TRUE_PARAMS` for GA, CMA-ES and multi-start (FD and AD), locked into the suite by `tests/test_benchmarks.py:249`.

One caveat worth recording rather than leaving implicit: the full-model cases in `test_UQ.py` are `@pytest.mark.manual`, so it is the emulator-based pair in `test_uq_on_emulator.py` that actually runs on every push. If you want the full-model recovery on CI too, that is a narrower follow-up than this issue.

PR #408, opened 2026-08-12
feat(uq): pyMC backend and the posterior-recovery tests (#195, #179, #367 step 6)

### LINKED ISSUE #195: Introducing pyMC
Incorporating pyMC library for uncertainty quantification (UQ) tools including MCMC, SMC (ABC), etc.

The `mcmc_options` entry in the `user_input.yaml` file could change to `UQ_options` with sub-entries for `library` (emcee/zeus/pymc) and `method` (MCMC, SMC, etc)

FinbarArgus:
Closes #195. Last of the stack bringing in #367 (pyMC for UQ). Original work by @MOH-Shafizadegan.

> Stacked on #402–#406 — their commits are included here and the diff shrinks to the last commit once those merge.

## What

emcee's affine-invariant ensemble sampler is a good default, but it is one algorithm. pyMC brings Metropolis and sequential Monte Carlo, and SMC in particular can cross the low-probability regions between separated modes that an ensemble sampler gets stuck on.

`PyMCSampler` wears emcee's interface — `run_mcmc` / `get_chain` returning a `(steps, walkers, params)` chain — so `UQ_options['library']` changes one line of construction and **nothing downstream**: the sampling loop, `mcmc_chain.npy`, the corner plots and the R-hat diagnostics are all unchanged.

This also retires the module-level `mcmc_lib = 'emcee'  # TODO make this a user variable` constant, which meant editing the source to change sampler.

## Optional dependency, never at module level

`pymc`/`pytensor` are imported inside the backend module and added as an optional `[uq]` extra. `paramID.py` is imported by every calibration run, so #367's top-level `import pytensor` would have broken every user who has not installed it. Selecting the backend without it raises a message naming `pip install -e ".[uq]"` **and** the way to carry on without it.

## Three corrections to the original

**1. The prior was counted twice.** #367's pyMC model declared `pm.Uniform` / `pm.Exponential` / `pm.Normal` priors *and* put `get_lnlikelihood_lnprior_from_params` — which already includes the log prior — in the `Potential`. The sampled posterior was therefore not the one CA defines.

Here the pyMC variables are bare *support* declarations (`Uniform` over the box, or `Flat` when the parameter is unbounded), which contribute only a constant inside that support, and CA's log posterior is the single density term.

**2. Those re-declared priors hardcoded the old defaults** (`lambda=1`, `sigma=(max-min)/6`, `mu=(max+min)/2`), silently ignoring the `prior_mean` / `prior_std` / `prior_origin` / `prior_scale` hyper-parameters and the `unbounded` flag that `params_for_id` supports on current master. Deferring to CA's own `get_lnprior_from_params` means there is one implementation of the prior vocabulary rather than two that can drift.

**3. `num_walkers // num_procs` is 0 when there are more ranks than walkers**, so pyMC was asked for zero chains — an MPI run that quietly produces no samples. `chains_for_rank` floors at 1.

The trace conversion also raises on a trace missing its parameters instead of returning `None`, which became an unreadable failure far downstream after the sampling had already been paid for.

## Tests

**`tests/test_uq_pymc_backend.py`** — 12 fast unit tests covering everything that must be right *without* pymc installed: the `(chain, draw)` → `(steps, walkers, params)` conversion and that parameter order is preserved (position *is* a parameter's identity downstream); the missing-parameter and missing-posterior errors; the chain arithmetic including more ranks than walkers; the install message naming an extra that actually exists in `pyproject.toml`; emcee still being the default; an unknown library naming the valid ones; and that both backends expose the same surface.

**`tests/test_UQ.py`** — the three posterior-recovery tests (#179), adapted to the flat `UQ_options` and the new stats API.

`test_mcmc_unimodal_with_validation` **passes**: emcee recovers the analytical solution `[1.0, 1.0]` with the spread the observation noise implies, `r_hat < 1.3` and ESS > 50. That is the end-to-end validation of the whole stack — the config surface, the equal-weight likelihood and the diagnostics — not just of this PR.

`test_mcmc_unimodal_with_validation_KDE_likelihood` uses `library: pymc` and skips where the extra is absent.

Verified `not slow and not compare_optimisers`: **658 passed, 1 failed** — `test_python_BDF_solver[SN_simple]`, the known libCellML Analyser limitation (#151), which fails identically on master.

## ⚠️ What I could not verify

**pymc is not installed in the OpenCOR pythonshell this repo runs under, and I did not install it.** So the pyMC sampling path itself — `_build_model`, `pm.sample`, `pm.sample_smc` — has not been executed. Its adapter layer is tested, its dispatch is tested, and its failure mode without the dependency is tested, but the first real pyMC run will be someone's first real pyMC run. Whether `pytensor` (which compiles C extensions) installs cleanly into that interpreter is the open question flagged when this work was planned; it is deliberately last in the stack so it blocks nothing else.

## Not included from #367

The three seaborn posterior-predictive plots (`plot_boxplots_for_predictions`, `plot_distribution_grid`, `get_prior_pdf`) — see #406 for why they are better decided alongside #341.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -167,7 +167,8 @@ jobs:
   #
   # Measured on Simple_ODE_Benchmark, which is why the job is split the way it is:
   #
-  #   tests/test_uq_on_emulator.py                        ~1 min    (sampled through an emulator)
+  #   tests/test_uq_on_emulator.py [emcee]                ~34 s    (sampled through an emulator)
+  #   tests/test_uq_on_emulator.py [pymc]                 ~83 s    (same, pyMC backend)
   #   test_UQ.py::test_mcmc_unimodal_with_validation      ~9 min    (emcee, real model)
   #   test_UQ.py::test_mcmc_bimodal_with_validation      ~26 min    (emcee, real model)
   #   test_UQ.py::..._KDE_likelihood                     >50 min    (pyMC -- killed, unfinished)
@@ -176,7 +177,8 @@ jobs:
   # likelihood evaluation *per parameter per step* where emcee moves the whole ensemble at once,
   # and PyMCSampler's num_tune defaults to 1000, so it runs 1500 iterations for the 500 draws
   # the test asks for. Run it by hand (or fix the tuning cost) rather than putting an hour on
-  # every pull request; the pyMC backend's other coverage is in tests/test_uq_pymc_backend.py.
+  # every pull request. The pyMC backend is still covered end to end -- through the emulator,
+  # where the same chain costs 83 seconds instead of an hour.
   #
   # It is its own job rather than extra cases on test-param-id-*, because it needs two optional
   # extras those jobs have no use for, and because at ~35 minutes it would become the critical
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -161,15 +161,27 @@ jobs:
       run: |
         python -m pytest tests/test_param_id.py tests/test_sensitivity_analysis.py -v --with-mpi --tb=short --junitxml=junit-param-id-serial.xml
 
-  # The end-to-end UQ tests: a real calibration followed by a real MCMC chain, checked against
-  # a posterior whose answer is known in closed form (issue #179). They are marked `slow` and so
-  # are deselected everywhere else -- this is the only job that runs them.
+  # The end-to-end UQ tests: a real chain checked against a posterior whose answer is known in
+  # closed form (issue #179). Nothing else runs them -- the full-model ones are marked `slow`
+  # and every other job deselects them.
   #
-  # It is its own job, not extra cases bolted onto test-param-id-*, for two reasons: it needs the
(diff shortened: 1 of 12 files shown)


FinbarArgus:
Update on verification: **`test_mcmc_bimodal_with_validation` also passes** (1578s).

That is the more demanding of the two emcee recovery tests — the chain has to represent both modes rather than parking every walker in whichever one it found first, which is precisely the failure a point estimate cannot even express. Both emcee posterior-recovery tests in `tests/test_UQ.py` are therefore green:

| test | sampler | result |
|---|---|---|
| `test_mcmc_unimodal_with_validation` | emcee | **passed** (534s) — recovers [1.0, 1.0], r_hat < 1.3, ESS > 50 |
| `test_mcmc_bimodal_with_validation` | emcee | **passed** (1578s) — both modes represented |
| `test_mcmc_unimodal_with_validation_KDE_likelihood` | pymc | skipped — needs the optional `[uq]` extra |

The caveat in the description stands unchanged: the pyMC sampling path itself is still unexecuted here, because pymc is not installed in the OpenCOR pythonshell.

Issue #84, opened 2025-05-19
Include kwargs for user defined cost functions


FinbarArgus:
User defined cost functions should have kwargs for weights, std, and whatever the user wants to define. Currently weights and stds are required entries for specific cost functions but we need to make these general.

PR #423, opened 2026-08-13
A distribution is a ground truth, not a data_type (#421)

### LINKED ISSUE #421: prob_dist is not a data_type: all data_types have a prob_dist, changing functionality of kde estimator cost function.
`prob_dist` is a fourth `data_type` alongside `constant`, `series` and `frequency`. The other three describe the **shape of the simulated data**: a scalar, a trace, an FFT. `prob_dist` does not — it describes the shape of the **ground truth**, and the feature it produces is an ordinary scalar.

## The feature is produced identically to a constant

`get_obs_output_dict` files the *same* `obs` in both branches:

```python
if data_types[JJ] == 'constant':
    obs_const_vec[const_count] = obs
    const_count += 1
...
elif data_types[JJ] == 'prob_dist':
    obs_val_for_prob_dist_vec[prob_dist_count] = obs
    prob_dist_count += 1
```

Same operation, same operands, same scalar. `obs_val_for_prob_dist_vec` is a 1-D array of scalars, never an array per item. The only difference is which vector it lands in.

## The real difference is which ground truth the cost receives

`value`/`std` versus `prob_dist_params`. CA already decides that kind of thing from the cost func's signature — `framework_kwargs_for` says so itself:

> `gaussian_MLE(output, desired_mean, std, weight)` gets both; `multimodal_gaussian(output, prob_dist_params, weight)` gets only `weight`, because it has nowhere to put a std -- which is the point of this function, and previously meant it could not share a call site.

#84 generalised the **keyword** arguments across cost shapes. The **positional** ground truth stayed hardcoded, and which one an observable gets is chosen by which cost loop it is in — i.e. by its `data_type`. The generalisation was done halfway.

## What the split costs

Not tidiness:

- **Scalar KDE observables are invisible to the emulator and to FD sensitivities.** `observable_features` and `emulated_feature_labels` index by `const_idx_to_obs_idx`, so a `prob_dist` item has no slot. Training an emulator on an obs_data whose items are all `prob_dist` fails outright:
  `ValueError: the emulator predicts scalar data_item features only, but obs_data.json has data_type(s) ['prob_dist']`.
  That is what blocks emulating the bimodal UQ test (`tests/test_UQ.py::test_mcmc_bimodal_with_validation`, ~26 min against the real model), whose bimodality lives entirely in the KDE likelihood while the model response is smooth — an easy emulator problem that cannot currently be posed.
- **Parallel machinery for no gain**: its own weight vector (`scaled_weight_prob_dist_from_exp_sub`), index map (`prob_dist_idx_to_obs_idx`), cost loop, casadi branch and plotting branch — ~67 references across 6 files.
- **Dead fields that look meaningful.** `tests/test_inputs/Simple_ODE_Benchmark_bimodal_obs_data.json` sets `"value": 1, "std": 0.1` on items whose cost (`kernel_density_estimation`) never reads either.

## Proposed

Remove `prob_dist` as a `data_type`. An observable compared against a distribution becomes:

```json
{
  "data_type": "constant",
  "operation": "steady_state_avg",
  "operands": ["benchmark/x"],
  "cost_type": "kernel_density_estimation",
  "prob_dist_params": { "data_points": [...], "bandwidth": 0.1 }
}
```

and the cost loop picks the ground truth from the cost func's second positional parameter — `prob_dist_params` for KDE and `multimodal_gaussian`, `desired_mean` for `gaussian_MLE` — extending the rule #84 already established for keyword arguments.

Then a distribution-compared observable is a scalar like any other, and everything scalar-shaped works on it for free: the emulator, FD sensitivities, `observable_features`.

`series` and `frequency` stay as they are; they are genuinely different data shapes.

FinbarArgus:
Closes #421.

`prob_dist` was a fourth `data_type` beside `constant`, `series` and `frequency`. The other three describe the shape of the **simulated data** — a scalar, a trace, an FFT. `prob_dist` did not: it described the shape of the **ground truth**, while the feature it produced was an ordinary scalar. `get_obs_output_dict` filed the *same* `obs` in both branches; only the vector it landed in differed.

That split was not merely untidy. Everything that works on scalar features indexes by `const_idx_to_obs_idx`, so a `prob_dist` item had no slot in it and was invisible to the emulator and to FD sensitivities. Training an emulator on such an obs_data failed outright:

```
ValueError: the emulator predicts scalar data_item features only, but obs_data.json
has data_type(s) ['prob_dist'] at data_item index(es) [0, 1]
```

which is what blocked emulating the bimodal UQ case — the one whose bimodality lives entirely in the KDE likelihood while the model response is smooth, i.e. an easy emulator problem that could not be posed.

## What changes

An observable compared against a distribution is now a **`constant`** that states its ground truth as `prob_dist_params` instead of `value`/`std`:

```json
{
  "variable": "x_steady_state",
  "data_type": "constant",
  "operation": "steady_state_avg",
  "operands": ["benchmark/x"],
  "unit": "dimensionless",
  "weight": 1.0,
  "cost_type": "kernel_density_estimation",
  "prob_dist_params": {"data_points": [1.05, 0.99, 4.08, "..."]},
  "cost_kwargs": {"bandwidth": 0.1}
}
```

**Which ground truth a cost receives is read off its signature** — the second positional parameter: `desired_mean` for `gaussian_MLE`/`MSE`/`AE`, `prob_dist_params` for `kernel_density_estimation`/`multimodal_gaussian`/`poisson_MLE`. #84 generalised the *keyword* arguments across cost shapes; the positional ground truth stayed hardcoded, chosen by which cost loop an item was in, i.e. by its `data_type`. This extends the same rule to that slot, so there is one dispatch rule rather than two.

**`bandwidth` moves into `cost_kwargs`.** It tunes the comparison rather than stating the measurements, so it belongs where the other knobs live: it can now be swept without editing the data, and a typo in it is rejected by the #84 validation instead of silently ignored.

**`value`/`std` are required *unless* the item sets `prob_dist_params`.** The fixtures used to carry `"value": 1, "std": 0.1` next to a `kernel_density_estimation` cost that read neither. An item with *neither* ground truth is now rejected at parse time rather than contributing a `nan` that propagates and looks like a solver failure.

Removed with the data_type: `scaled_weight_prob_dist_from_exp_sub`, `prob_dist_idx_to_obs_idx`, `val_for_prob_dist`, the `prob_dist` cost loop, and the casadi and plotting branches.

## Details worth a look in review

- **Indexing.** `ground_truth_prob_dist_params` is full length over all data_items and read by row, like `cost_type` and the weight vectors — not by a compacted counter of its own. Disagreeing index spaces is exactly the #349 bug.
- **Plots.** A distribution ground truth is drawn as its modes (`multimodal_gaussian`), or as the 5th/50th/95th percentiles of the samples (KDE). Quantiles rather than a mean, because the mean of a bimodal sample sits where no measurement ever landed. Percent error is taken against the *nearest* reference value, so a fit that lands squarely on the other mode is not reported as a large error.
- **Gradients.** Symbolic AD now refuses a distribution cost *by name*: its density is built from numbers (scipy's `gaussian_kde`, a mixture) and cannot take a symbol. These items used to be refused as part of the `prob_dist` check; as constants they would otherwise have picked up the `nan` standing in for the `value` they deliberately do not have. AADC already refuses anything but `gaussian_MLE`/`MSE`; the FSA path reconstructs the cost numerically and is unaffected; the Laplace FIM already skips an observable without a positive std.
- **No backwards compatibility**, as agreed on the issue. An obs_data still using `data_type: "prob_dist"` is rejected at parse time with a message saying exactly what to write instead.

## Tests

`tests/test_distribution_ground_truth.py` — 19 new unit tests: the data_type is gone from `VALID_DATA_TYPES` (which CUFLynx reads for its menu), the old spelling is rejected with a useful message, a KDE item occupies a constant slot, the distribution is indexed by row, `value`/`std` are conditionally required, the ground truth is read off each cost's signature, and the KDE cost scores *both* modes as good through `cost_calc` end to end while the point between them costs more.

`tests/test_uq_on_emulator.py::test_uq_on_an_emulator_finds_both_modes_of_a_bimodal_target` — the case this unblocks, and it also asserts the training set spans both KDE clusters, since a mode outside the emulator's training box is unreachable however good the sampler is:

```
held-out R2   1.0000   x_{SS} (steady_state_avg benchmark/x)
held-out R2   1.0000   y_{SS} (steady_state_avg benchmark/y)
[bimodal] 40.0% of samples below the midpoint 2.559
1 passed in 62.14s
```

62 seconds, against ~26 minutes for the full-model equivalent (`test_UQ.py::test_mcmc_bimodal_with_validation`, marked `slow`, so it never runs on a PR).

Both KDE fixtures are converted, and the tutorial gains a short section on stating a ground truth as a distribution.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -186,7 +186,10 @@ jobs:
   # instead of adding to it.
   test-uq:
     runs-on: ubuntu-latest
-    timeout-minutes: 90
+    # ~5 minutes of tests plus ~3 of installs. Deliberately not generous: this job used to take
+    # 89 minutes against the real model, and a timeout close to the real cost is what makes a
+    # regression back to that visible instead of merely slow.
+    timeout-minutes: 25
     steps:
     - uses: actions/checkout@v4
 
@@ -208,17 +211,19 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y libopenmpi-dev openmpi-bin libsundials-dev build-essential
 
-    # Fast enough to be worth running first: it fails in a minute if UQ is broken at all,
-    # rather than after half an hour of sampling the real model.
-    - name: Run UQ on an emulator (fast posterior-recovery check)
+    # The whole UQ stack -- the cost, the priors, both samplers, the statistics -- checked
+    # against a posterior whose answer is known in closed form, including the bimodal case.
+    #
+    # The full-model equivalents in test_UQ.py are deliberately not run here. They took 82
+    # minutes of this job's 89 (the bimodal one alone: 56), and they sample the same posteriors
+    # these do -- the only difference being that the likelihood comes from the solver rather
+    # than from an emulator faithful to it (held-out R2 1.0000). Paying 16x the wall clock on
+    # every PR for that difference is not a good trade, so they are marked `manual` and run on
+    # demand with --run-manual when the solver-driven path itself changes.
+    - name: Run UQ on an emulator (posterior recovery, both samplers, bimodal)
       run: |
         python -m pytest tests/test_uq_on_emulator.py -v --with-mpi --tb=short \
           --durations=0 --junitxml=junit-uq-emulator.xml
-
-    - name: Run the full-model UQ posterior-recovery tests
-      run: |
-        python -m pytest tests/test_UQ.py -m slow -k "not KDE" -v --with-mpi --tb=short \
-          --durations=0 --junitxml=junit-uq.xml
 
     # A skipped test is a green step, so this job could report success having run nothing at
     # all -- and that is the likely failure, not a red X: every UQ test is guarded by a skipif
@@ -271,7 +276,6 @@ jobs:
         name: uq-test-results
         path: |
           junit-uq-emulator.xml
-          junit-uq.xml
 
   # Job to run full test suite if OpenCOR is available (optional)
   # This can be enabled if OpenCOR is set up in CI environment
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -138,7 +138,7 @@ A `params_to_change` value is a **float** (constant) or a **string** (trace key 
 
 ## Backend caveats
 
-- **SN_simple / SN_full**: `cellml_only` generation + **Myokit** accept the emitted CellML (including state initial values referencing `*_init` params). **`PythonGenerator`** uses libCellML Analyser and requires a strict **ODE** model; the same SN CellML fails `ANALYSER_VARIABLE_NON_CONSTANT_INITIALISATION`, so `model_type: python` codegen is **not expected to work for SN** until the generator/model satisfies the analyser.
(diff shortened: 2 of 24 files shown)


FinbarArgus:
Superseded by #427, which is the same work stacked on #425 and #424 so its CI runs on the split, two-rank jobs. No checks ever ran on this PR's head commit.

Issue #453, opened 2026-08-18
unable to access a constant variable in myokit solver


Jayakody148:
[link](https://github.com/physiomelinks/circulatory_autogen/blob/master/src/solver_wrappers/myokit_helper.py#L1233)
constant variables cannot be accessed within the last log. 
```python
def _extract(self, name):
        if name == "time":
            if self.last_log is None:
                raise RuntimeError("No log available")
            # Return the canonical cumulative time axis (tSim), not Myokit's logged
            # clock. Myokit's clock restarts at each sub-experiment's run_t0
            # (update_times -> reset), so the raw log is a per-segment local time and
            # corrupts time-based observables (e.g. first_peak_time) in multi-sub-
            # experiment protocols. tSim is built from start_time/pre_time in
            # _setup_time and is the same length as the logged state/var series, so it
            # stays aligned with the values get_results returns. This is identical to
            # get_time() and to the OpenCOR backend's time axis.
            return self.tSim - self.pre_time
        kind, qname = self._resolve_name(name)
        if self.last_log and kind in ("state", "var") :
            data = np.asarray(self.last_log[qname])
            return data
        if kind == "state":
            return np.asarray([self.simulation.state()[self.state_index[qname]]])
        if kind == "var":
            # if qname in self.default_values:
            #     return np.asarray([self.default_values[qname]])
            # Evaluate algebraic variable from current model state  
            if qname in self.qname_to_var:  
                try:  
                    current_value = self.qname_to_var[qname].eval()  
                    return np.asarray([current_value])  
                except Exception:  
                    pass  
        raise ValueError(f"variable {name} not found")
```

## Proposed solution

check whether the `qname` exists in the `last_log` :
```python
if self.last_log and kind in ("state", "var") and qname in self.last_log:
            data = np.asarray(self.last_log[qname])
            return data
```

PR #454, opened 2026-08-18
accessing a constant variable in myokit solver  

Jayakody148:
Fixing issue #453


FinbarArgus:
Thanks, great fix, i had refactored quite a lot so I brought your change into #455 and merged that.

Issue #461, opened 2026-08-19
Cache the apt packages: 106 min of runner time per push, and it has already failed a job


FinbarArgus:
Nine jobs across `tests.yml` (8) and `benchmarks.yml` (1) run the identical line:

```yaml
sudo apt-get update
sudo apt-get install -y libopenmpi-dev openmpi-bin libsundials-dev build-essential
```

Nothing caches it, so every job on every push re-downloads the same ~48-package dependency tree from `azure.archive.ubuntu.com`.

## What it costs

Measured on one healthy run of `master` (32207920981) — same packages, same push, same few minutes:

| Duration | Job |
|---|---|
| 22m35s | `test-param-id-serial` |
| 16m35s | `test-param-id-mpi` |
| 16m13s | `test-uq-mpi` |
| 13m19s | `test (3.11)` |
| 11m47s | `test-param-id-mpi-full-scale` |
| 11m20s | `test-param-id-serial-full-scale` |
| 10m14s | `test (3.9)` |
| 2m15s | `test-uq` |
| **1m22s** | `test (3.10)` |

**106 minutes of runner time per push**, to install four packages — and a **17x spread** between the fastest and slowest job in that same run. The 1m22s case is the useful number: when the mirror cooperates, this work is about ninety seconds. Everything above that is queueing on a slow mirror.

## Why it matters beyond the wall clock

It has already caused a failure that looks like a test failure and is not. On #460, `test-uq` reported **cancelled** with every test step skipped:

```
Install dependencies                        2m34s   ok
Install MPI and Myokit build dependencies  22m34s   <- killed at timeout-minutes: 25
Run the pyMC live-chain tests                       skipped
Run the multi-rank UQ tests                         skipped
Run UQ on an emulator                               skipped
```

The apt log shows the mirror degrading mid-step — `apt-get update` itself managed 485 kB/s, then package downloads collapsed:

```
Get:7  cmake         [11.2 MB]  02:43:19 -> 02:50:59   7m40s, ~24 kB/s
Ign:13 libsource-highlight4t64                          fetch failed, dropped
Get:14 gdb           [4.1 MB]   02:53:47 -> 02:56:09
Get:18 libamd-comgr2 [14.4 MB]  02:57:13 -> never finished
```

[...]

PR #462, opened 2026-08-19
ci: go straight to an apt mirror that answers, from one place instead of nine

FinbarArgus:
The identical apt step took **1m22s** in `test (3.10)` and **22m35s** in `test-param-id-serial` — in the *same run* of this workflow, same packages, same push. That 17x spread is not load. `azure.archive.ubuntu.com` stalls, and apt does not fail over quickly.

## What is actually happening

The runner image points apt at a mirrorlist (`/etc/apt/apt-mirrors.txt`) listing `azure.archive.ubuntu.com` first and `archive.ubuntu.com` as fallback. When Azure is unhealthy, apt spends its connect timeout plus retries on the dead host for *every* index and *every* package.

CUFLynx's release build showed the extreme version of this — a 32-second step hanging for 1h40m, then 25 minutes more on the rerun:

| Host | Result in that job |
|---|---|
| `azure.archive.ubuntu.com` | **22 × `Ign`, 0 successes** |
| `archive.ubuntu.com` | 1 Hit, 3 Get ✓ |
| `packages.microsoft.com` | 7 Get ✓ |
| `dl.google.com` | 2 Get ✓ |

Here it shows up less dramatically but more often: `test-uq` cancelled at its 25-minute ceiling **having never run a test**, and `test (3.11)` sitting 45 minutes in the same step.

## Nine copies became one

The step was copy-pasted nine times — eight in `tests.yml`, one in `benchmarks.yml` — so the fix would have been copy-pasted nine times too. It is now a composite action, which is what #461 asked for. Net **-36/+9** lines in the workflows.

Two details in it are deliberate:

**It probes rather than dropping Azure unconditionally.** When healthy it is the faster mirror — same datacentre — so it is worth keeping on a good day.

**The probe is `curl --max-time`, not a reachability check.** The failure mode is a *stalled transfer*, not a refused connection. Probing it by hand while it was down returned **13287 of 270087 bytes and then stopped** — the socket opens and data starts flowing, so anything that only tests reachability is fooled. That is precisely why apt sits there rather than failing over.

It also asks for the runner's own codename via `lsb_release -cs` rather than hardcoding one: `ubuntu-latest` is noble now and was jammy recently, and a hardcoded codename would silently probe the wrong path and always "succeed".

## Also here

`test` and `test-with-opencor` had no `timeout-minutes` — every other job in the file does. `test` is the one that mattered: it is where 3.11 could hang indefinitely, since without a value a job runs to GitHub's 360-minute default.

## Relationship to #461

**Not a substitute.** Caching is still worth doing for the 106-minutes-per-push it saves, but it would not have prevented any of the above: a cache miss — a dependency bump, a new runner image — walks into the same stall. Three layers, each catching what the previous misses: go to a working mirror → fail a dead one fast (`Acquire::*::Timeout 15`) → job timeout.

## Verified

Both workflows and the action parse; all nine call sites resolved; every job that uses the local action runs `actions/checkout` first (a `./` action is unavailable before checkout). The probe logic was exercised against the live outage — it correctly detects Azure as unhealthy today.

### DIFF
--- a/.github/actions/install-system-deps/action.yml
+++ b/.github/actions/install-system-deps/action.yml
@@ -1,0 +1,46 @@
+name: Install system libraries (MPI, SUNDIALS)
+description: >-
+  Install the OpenMPI/SUNDIALS/toolchain packages every Linux job here needs, going
+  straight to an Ubuntu mirror that answers.
+
+runs:
+  using: composite
+  steps:
+    - name: Drop the Azure apt mirror if it is not answering
+      shell: bash
+      run: |
+        # The runner image points apt at a mirrorlist (/etc/apt/apt-mirrors.txt) listing
+        # azure.archive.ubuntu.com first and archive.ubuntu.com as fallback. When the Azure
+        # mirror is down, apt does not fail over quickly: it spends its connect timeout plus
+        # retries on the dead host for *every* index and *every* package.
+        #
+        # That is not hypothetical. It hung CUFLynx's release build for 1h40m in a step whose
+        # healthy cost is 32 seconds, and it is the best explanation for the spread here --
+        # in one run of this workflow the identical step took 1m22s in `test (3.10)` and
+        # 22m35s in `test-param-id-serial`, which is what winning or losing the mirror
+        # lottery looks like.
+        #
+        # Probe rather than delete unconditionally: when Azure is healthy it is the faster
+        # mirror (it is in the same datacentre), so it is worth keeping on a good day.
+        #
+        # --max-time, not a connect check: the failure mode is a *stalled transfer*, not a
+        # refused connection. Probing it by hand while it was down returned 13287 of 270087
+        # bytes and then stopped -- the socket opens and data starts flowing, so anything that
+        # only tests reachability is fooled, which is precisely why apt sits there.
+        # Ask for this runner's own release: ubuntu-latest is noble now and was jammy not
+        # long ago, and a hardcoded codename would silently probe the wrong thing.
+        codename="$(lsb_release -cs)"
+        if ! curl -fsS --max-time 10 -o /dev/null \
+             "http://azure.archive.ubuntu.com/ubuntu/dists/${codename}/InRelease"; then
+          echo "azure.archive.ubuntu.com is not answering -- using archive.ubuntu.com"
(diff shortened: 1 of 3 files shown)


FinbarArgus:
**Correction — the probe in the first commit did not work, and this branch's own CI is what showed it.**

I reported "2.8 min across five jobs" earlier. That was a biased sample: those five were simply the jobs that finished first, and they were the fast ones. All nine are now in:

| Dependency step | Job |
|---|---|
| **30m33s** | `test-param-id-mpi` ← *worse* than the 22m35s it was meant to fix |
| 21m50s | `test-param-id-serial-full-scale` |
| 16m58s | `test-uq-mpi` |
| 13m52s | `test-param-id-serial` |
| 0m51s | `test-uq` |
| 0m34s | `test (3.10)` |
| 0m29s | `test (3.11)` |
| 0m28s | `test-param-id-mpi-full-scale` |
| 0m26s | `test (3.9)` |

**86 minutes against the old 106** — a 19% saving, not a fix. The run passed, and `test-uq` did go from cancelled-at-25m to 8m53s, but four jobs were untouched.

## Why

A threshold I got wrong. The probe allowed `curl --max-time 10`, and Azure answers in ~8–9 seconds for a 270 KB file — about **30 kB/s**. So the probe *passed*, the degraded mirror was kept, and apt crawled through 48 packages at that rate. A healthy mirror returns that file effectively instantly.

The deeper error: the failure mode is not "down", it is **degraded** — ~30 kB/s with roughly one request in three stalling outright. Any time budget loose enough to tolerate a healthy mirror's jitter also accepts one two orders of magnitude too slow. A timeout cannot tell "slow" from "working", and slow is the case that costs us.

## The fix

The optimisation goes — `ff2580b` drops the mirror unconditionally, matching CUFLynx #281, which is the version with clean numbers: a 32-second step that had hung for 1h40m completed in **50 seconds**, and the full release path built four binaries and passed four analysis-e2e runs in **14 minutes** total.

If "keep the faster mirror on a good day" is ever wanted back it needs a *throughput* floor (`curl --speed-limit`/`--speed-time`), not a timeout. That is written into the file so the next person does not re-derive it.

I will report the numbers across **all ten** jobs this time, not the ones that finish first.

PR #469, opened 2026-08-23
Let every stage command take --user-inputs PATH

FinbarArgus:
## The problem

Every console command reads its configuration from `user_run_files/user_inputs.yaml`, or from whatever `user_inputs_path_override` inside that file points at, under a user directory chosen by `$CUFLYNX_USER_DIR`. `_cli.py` said as much: *"none of them take options"*.

That makes a checkout hold exactly one configuration at a time. Switching between two — a training run and a per-dataset run, or one `obs_data.json` and another — means editing a file in place, hand-editing a redirect key, or moving an environment variable. None of those is something a script can do safely while another rank is reading the file.

## What this does

The engine already supported the alternative:

- `parse_user_inputs_file` only opens the default file when `inp_data_dict is None`;
- every stage's `run_*` function already takes `inp_data_dict` as its first parameter.

So this adds the CLI surface and nothing else. `--user-inputs PATH` loads the named yaml, and the stage passes it to the function it was already calling.

Absent, the value is `None` — which is exactly what the stages already treat as *"read the configured file yourself"*. **Behaviour with no option is unchanged.**

```
mpiexec -n 4 cuflynx-param-id --user-inputs /path/to/ox1_user_inputs.yaml
cuflynx-train-emulator --user-inputs /path/to/full_user_inputs.yaml
```

Failures are reported by name at parse time rather than from inside a run: a missing file, and a yaml that parses to a list or a scalar instead of a mapping of settings.

Every rank reads the file. That matches how the default file is already loaded; broadcasting would only move the same failure to a later collective.

## Tests

`tests/test_console_entry_points.py` is table-driven off `[project.scripts]`, so all seven commands are covered automatically:

- each offers `--user-inputs` in `--help`;
- each reports a missing file by name, non-zero, before doing any work.

`load_user_inputs` is covered directly for the three cases that decide behaviour — absent, a real file, and a non-mapping.

124 passed (up from 107).

## Checked by hand

Driving a real generation from an out-of-tree yaml:

```
python -m libcuflynx.scripts.script_generate_with_new_architecture \
    --user-inputs .../SN_full/resources/ox1_user_inputs.yaml
# Model generation has been successful. Validation status: Myokit run succeeded.
```

### DIFF
--- a/src/libcuflynx/scripts/_cli.py
+++ b/src/libcuflynx/scripts/_cli.py
@@ -17,13 +17,17 @@ Two things every one of them needs, defined once here:
   killed by hand.
 """
 import argparse
+import os
 import traceback
+
+import yaml
 
 #: Every stage takes its configuration from the same file, and none of them take
 #: options. Saying so in ``--help`` is the whole of the help text people need.
 CONFIG_EPILOG = (
-    "Configuration is read from user_run_files/user_inputs.yaml under the user directory,\n"
-    "or from the file named by user_inputs_path_override in it.\n"
+    "Configuration is read from the file named by --user-inputs; otherwise from\n"
+    "user_run_files/user_inputs.yaml under the user directory, or from the file named by\n"
+    "user_inputs_path_override in it.\n"
     "\n"
     "The user directory is $CUFLYNX_USER_DIR if set; otherwise the circulatory_autogen\n"
     "checkout this libcuflynx was run from, if it is one; otherwise the current directory.\n"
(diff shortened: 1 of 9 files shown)


FinbarArgus:
`test-uq` and `test-uq-mpi` fail here at **dependency install**, in ~20s, before any test runs:

```
```

Not from this change — Cython 3.3.0 was published on 2026-08-22 and made that assignment an error. `test-uq` passed on `master` two days ago and the same failure appears on every branch that installs the `emulation` extra.

#472 fixes it (a `--build-constraint` ceiling, mirroring what CUFLynx already did). Once that is in, this should go green without a change here.

PR #470, opened 2026-08-23
Write a runnable pipeline bundle from a user_inputs.yaml

FinbarArgus:
> Stacked on #469 (`--user-inputs`). Review that first; this PR's base will move to `master` once it merges.

## Why

A configuration is not a run. `user_inputs.yaml` says what to do, but reproducing it later — on another machine, in a paper's supplementary material, six months on — also needs the model, the observations, the parameter bounds, and something to drive them in the right order.

CUFLynx already knew how to assemble that folder, and the scripts it writes drive *this* engine. Keeping them in the GUI means the engine can change under a bundle nobody can regenerate without the GUI installed, and it means a bundle is only reachable by clicking Export. So the bundling moves here; CUFLynx keeps the part that is genuinely its own — turning GUI settings into a `user_inputs` dict — and will call this to write the folder.

## Using it

```
cuflynx-generate-pipeline --user-inputs my_run.yaml --output-dir ./bundle
cd bundle && python run_pipeline.py
```

```
user_inputs_<yymmdd>.yaml    the configuration, every path relative
run_pipeline.py              the driver; each stage gated by a do_* flag
generated_models/<prefix>/   the CellML model
resources/                   obs_data.json, params_for_id.csv, any user funcs
emulator/                    a trained emulator, when one is in use
```

## The bundle is movable

That is the whole point, and what the tests are about: every input is copied in, every path in the written yaml is relative to the folder, and the absolute directories describing the machine it came from are dropped — `run_pipeline.py` rebuilds them from its own location. A bundle that only runs where it was made looks identical until someone unpacks it somewhere else.

`run_pipeline.py` is static. Everything specific to a study is in the yaml beside it, so there is no generated code to audit, and it finds its own configuration (the newest `user_inputs_*.yaml` next to it).

## Two things the exported pipeline never had

**An emulator stage.** `do_emulation` trains a surrogate before the analyses, since `use_emulator` makes them evaluate it instead of the solver. It refuses to continue when autoemulate fits nothing — that returns rather than raising, so an unchecked run reported success having trained nothing, and the next stage loaded a bundle that was not there. The message names `num_train_samples`, which is almost always the cause.

**Emulator settings reaching `inp_data_dict`.** `build_inp_data_dict` now copies `do_emulation`/`use_emulator`/`emulator_settings`, which nothing did before there was a stage to configure, and resolves a relative `emulator_dir` against the bundle. If `use_emulator` is set and there is no emulator to carry, writing the bundle says so rather than leaving an absolute path that resolves on one machine only.

## Tests

`tests/test_generate_pipeline_script.py` — 19 tests: bundle contents, relativised paths, dropped machine paths, the three refusals (missing model, missing `file_prefix`, unlocatable model), emulator carried / warned about, and the generated script parsed and checked for each stage gate. One runs the written bundle end to end with every stage off, which is what proves the yaml it writes is the yaml it can read.

`tests/test_console_entry_points.py` picks the new command up automatically (table-driven off `[project.scripts]`). 130 + 19 passing.

## Deliberately not moved

The plotting scripts (`PLOT_UTILITIES_SCRIPT`, `render_plotting_script`) stay in CUFLynx for now — ~900 lines of matplotlib coupled to the GUI's obs_data record shape, and not needed to run a pipeline.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,6 +34,7 @@ Runs are launched via shell scripts in `user_run_files/`. Each one invokes a **c
 | `run_identifiability_analysis.sh` | `cuflynx-identifiability` → `identifiability_run_script` | Laplace / profile-likelihood |
 | `run_emulator_training.sh` (arg: `num_processors`, uses `mpiexec`) | `cuflynx-train-emulator` → `train_emulator_run_script` | Train a surrogate of the obs features |
 | `plot_param_id.sh` | `cuflynx-plot` → `plot_param_id_script` | Plot calibration results |
+| _(no run script)_ | `cuflynx-generate-pipeline` → `generate_pipeline_script` | Write a self-contained bundle — model, resources, a `user_inputs_*.yaml` with every path relative, and a `run_pipeline.py` — that reruns the study elsewhere. Not driven by a `user_run_files/*.sh`: it takes `--user-inputs` and `--output-dir` directly. |
 
 Every command takes `--help` and no options beyond the optional `True|False` the two generation launchers pass; the configuration is the yaml. Each is a `main()` in the named module — `tests/test_console_entry_points.py` is table-driven off `[project.scripts]`, so a new entry point cannot be added without being tested.
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Each pipeline stage also has a console command, all configured from
 | `cuflynx-identifiability` | Laplace / profile-likelihood identifiability |
 | `cuflynx-train-emulator` | train a surrogate of the obs features |
 | `cuflynx-plot` | plot calibration results |
+| `cuflynx-generate-pipeline` | write a self-contained, movable bundle that reruns the study |
 
 ### `CUFLYNX_USER_DIR` — where an installed libcuflynx reads and writes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,10 @@ cuflynx-sensitivity = "libcuflynx.scripts.sensitivity_analysis_run_script:main"
 cuflynx-identifiability = "libcuflynx.scripts.identifiability_run_script:main"
 cuflynx-train-emulator = "libcuflynx.scripts.train_emulator_run_script:main"
 cuflynx-plot = "libcuflynx.scripts.plot_param_id_script:main"
+# Writes a folder that reproduces a configuration: the model, the observations, the
+# parameter bounds, the config, and a run_pipeline.py that drives every stage its do_*
+# flags enable. CUFLynx exports the same bundle from the GUI and calls this to write it.
+cuflynx-generate-pipeline = "libcuflynx.scripts.generate_pipeline_script:main"
 
 [tool.setuptools]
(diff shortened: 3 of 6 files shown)


FinbarArgus:
Pushed a correction from self-review (bca9f0c), and the PR description above is now wrong in one place — flagging rather than quietly editing it.

**What I got wrong.** The emulator stage guarded `if bundle is None: sys.exit(...)`, justified as "autoemulate can fail every candidate and still return". That is not true: a failure to fit raises out of `AutoEmulate.best_result()`. There was never a silent-success case to guard.

Worse, the guard was itself a bug. `EmulatorTrainer.train()` returns the bundle on rank 0 and `None` everywhere else — only rank 0 fits — so under `mpiexec` every non-root rank would have concluded nothing was trained, called `sys.exit`, and torn the job down while rank 0 was still fitting. `bundle.error_stats()` would have raised on them regardless.

I hit exactly this in my own runner on a 10-rank run: nine ranks printed "emulator training produced no model" for a run that was fine.

**Now:** reporting is gated on `getattr(trainer, "rank", 0) == 0`, matching the idiom the rest of the generated script already uses for MCMC and identifiability, and the bogus exit is gone. `min_r2` is what actually refuses a weak emulator, at use time.

A test asserts the stage gates on rank 0 and contains no `sys.exit`. 20 passing.

Also: this branch gets no CI because the workflow is filtered to `pull_request: branches: [master, main, devel]` and this is stacked on #469. I dispatched a `workflow_dispatch` run against the branch so it has coverage.

FinbarArgus:
**Shipped in master via the integration PR #477 — closing.**

Every commit on this branch is in `master`:

- `63407f3` Write a runnable pipeline bundle from a user_inputs.yaml
- `d7d3c76` docs: document cuflynx-generate-pipeline

merged through `b0b8193` on #477. `master` carries `cuflynx-generate-pipeline` in
`[project.scripts]` and documents it in both `README.md` and `CLAUDE.md`, and
`git merge-base --is-ancestor` confirms this head is an ancestor of `master`.

GitHub did not close this automatically the way it closed #469 and #471–#476, because
those seven were based on `master` and this one is based on
`feat/user-inputs-cli-option` (#469's branch). A PR is auto-closed when its head becomes
reachable from **its own base**, and that branch has not moved — so nothing ever pointed
GitHub at `master`.

Retargeting the base to `master` was the tidier fix, and GitHub refuses it for the same
reason this is safe to close: *"There are no new commits between base branch 'master' and
head branch 'feat/pipeline-script-generator'"*.

Two things worth carrying forward, both consequences of the stacked base:

- **This PR never ran CI at all.** No workflow triggers on a PR whose base is another
  topic branch. That is how it reached the integration branch adding
  `cuflynx-generate-pipeline` to `[project.scripts]` while documenting it nowhere —
  `test_docs_consistency` had been on `master` all along and would have caught it on the
  first push. Worth basing the next stacked PR on `master` and merging the parent first,
  or accepting that it is unverified until integration.
- `feat/user-inputs-cli-option` can be deleted now that #469 has merged.

PR #471, opened 2026-08-23
Give the sensitivity analysis's sim helper its protocol

FinbarArgus:
## The bug

A `params_to_change` entry may name a protocol trace by string — that is how an AP-clamp or voltage-clamp study drives a variable from a measured waveform — and resolving that name against `protocol_info['protocol_traces']` is the simulation helper's job.

`sobol_SA` built its helper and **never called `set_protocol_info`**. `param_id` (`paramID.py:1955`, `:2182`) and `ProtocolRunner` (`protocol_runner.py:162`) both have, always.

So sensitivity analysis was the one path where a protocol-trace study could not run. It generated the whole Sobol design, started sampling, and failed on the first evaluation with:

```
ValueError: params_to_change entry is a string trace key, but protocol_traces
not found in protocol_info.
```

which names neither the stage that lost them nor the fact that every other stage would have worked on the same `obs_data.json`.

## The fix

One line, where the helper is built — alongside the `update_times` that was already there:

```python
self._sim_helper = self.initialise_sim_helper()
self._sim_helper.set_protocol_info(self.protocol_info)
```

## Tests

`tests/test_sa_protocol_info.py` — a unit test rather than an integration one: the bug is a missing call, and reaching it through a real model means generating one and running `num_samples*(2M+2)` simulations first.

Verified the tests actually catch it — with the fix reverted, 2 of the 4 fail:

```
FAILED test_sim_helper_is_given_the_protocol - AssertionError: assert None is {...}
FAILED test_the_traces_reach_the_helper - TypeError: argument of type 'NoneType' is not iterable
```

## Checked against a real study

A 22-experiment protocol whose first six experiments are Wistar voltage-clamp waveforms (`soma_SN/V_set` naming entries in `protocol_traces`). Before: raised on sample 1 of 38. After: runs Sobol to completion and writes `all_outputs_n1_Sobol_indices.csv` and the first/total-order heatmaps.

### DIFF
--- a/src/libcuflynx/sensitivity_analysis/sobolSA.py
+++ b/src/libcuflynx/sensitivity_analysis/sobolSA.py
@@ -305,6 +305,15 @@ class sobol_SA():
         """
         if self._sim_helper is None:
             self._sim_helper = self.initialise_sim_helper()
+            # The helper needs the protocol to run one: a params_to_change entry may name a
+            # protocol trace by string, and resolving that name is the helper's job. Without
+            # this, an obs_data whose protocol drives a variable from a measured waveform --
+            # any AP-clamp or voltage-clamp study -- fails on the first sample with
+            # "params_to_change entry is a string trace key, but protocol_traces not found in
+            # protocol_info", after the whole design has been generated. param_id and
+            # ProtocolRunner have always set it; sensitivity analysis was the one path that
+            # did not.
+            self._sim_helper.set_protocol_info(self.protocol_info)
             if self.sim_time is not None and self.pre_time is not None:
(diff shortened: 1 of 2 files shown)


FinbarArgus:
`test-uq` and `test-uq-mpi` fail here at **dependency install**, in ~20s, before any test runs:

```
```

Not from this change — Cython 3.3.0 was published on 2026-08-22 and made that assignment an error. `test-uq` passed on `master` two days ago and the same failure appears on every branch that installs the `emulation` extra.

#472 fixes it (a `--build-constraint` ceiling, mirroring what CUFLynx already did). Once that is in, this should go green without a change here.

PR #472, opened 2026-08-23
Hold Cython below 3.3 when pip has to build the emulation extra's sdists

FinbarArgus:
## What happened

`test-uq` and `test-uq-mpi` started failing **in twenty seconds, at dependency install**, on branches that had touched nothing related ([#469](https://github.com/physiomelinks/circulatory_autogen/pull/469), [#471](https://github.com/physiomelinks/circulatory_autogen/pull/471)). `test-uq` passed on master two days ago.

Cython 3.3.0 was published on 2026-08-22 and made an implicit `double -> long` assignment an error. `harmonic` — reached through `autoemulate`, so only via the `emulation` extra — ships wheels for cp311+ only, so the Python 3.10 jobs build its sdist, and harmonic 1.3.0's `model_legacy.pyx:433` does exactly that assignment:

```
Cython.Compiler.Errors.CompileError: harmonic/model_legacy.pyx
```

## The fix

harmonic's own `[build-system] requires` is an unpinned `Cython`, so the ceiling has to come from our side: `build-constraints.txt`, passed with `--build-constraint` to every workflow install that reaches the extra.

That constrains the *build* environments pip creates for sdists and nothing libcuflynx runs against — which is why it is not a `pyproject` pin.

Three install steps reach the extra (`test-uq`, `test-uq-mpi`, `test-fenics-emulator`); all three already upgrade pip first, which `--build-constraint` needs.

CUFLynx hit this first and [fixed it the same way](https://github.com/physiomelinks/CUFLynx/commit/afd01b9); the two ceilings are for the same reason and should go at the same time.

## Guards

`tests/test_build_constraints.py`, four of them:

- the ceiling is present, with a message saying when it may go;
- *some* job actually installs the extra — otherwise the sweep below proves nothing;
- every such install passes the constraints file;
- every step that passes it upgrades pip first — `--build-constraint` is a pip 25.1+ option, and passing it to the runner image's pip is the same red build by a different route.

The workflows are parsed as YAML rather than scanned: the property is "this step installs the extra **and this same step** passes the flag", and a regex cannot tell a step's script from the comment above the next one.

## Removing it

When harmonic ships a cp310 wheel or fixes the `.pyx` — not before, or every job installing the emulation extra goes red again. `build-constraints.txt` says so.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -271,7 +271,10 @@ jobs:
         # [uq] brings pymc/pytensor and [emulation] brings autoemulate. Without them the
         # corresponding tests skip, and those code paths go unexercised in CI -- which is how
         # the pyMC backend stayed unproven for as long as it did.
-        pip install -e ".[dev,uq,emulation]"
+        # --build-constraint, not a runtime pin: [emulation] reaches harmonic, which has no
+        # cp310 wheel and so is built here. See build-constraints.txt for what it holds back.
+        pip install --build-constraint "${{ github.workspace }}/build-constraints.txt" \
+          -e ".[dev,uq,emulation]"
 
     - uses: ./.github/actions/install-system-deps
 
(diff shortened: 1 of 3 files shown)


FinbarArgus:
Confirmed on this run:

| job | before | after |
|---|---|---|
| `test-uq` | fail, 20s (at dependency install) | **pass, 8m56s** |
| `test-uq-mpi` | fail, 20s | **pass, 8m16s** |
| `test-fenics-emulator` | fail | **pass, 2m50s** |

All three are the jobs that install the `emulation` extra. The 20s failures were pip building `harmonic`'s sdist; with the ceiling they build and the tests actually run.

PR #479, opened 2026-08-23
Move the plotting scripts here too, and draw the posterior

FinbarArgus:
> Stacked on #470. Reads what #478 writes, but degrades quietly without it.

## Why

#470 deliberately left the plotting behind — ~1400 lines of matplotlib coupled to CUFLynx's `obs_data` shape, and not needed to *run* a pipeline. But a folder that reproduces a study and cannot draw it is half a bundle, and these figures describe **this engine's** output files, so they belong with the engine for the same reason `run_pipeline.py` does.

`plot_utilities.py` (the machinery) and `plot_outputs.py` (the half a user edits) are now written into every bundle.

## Three panels nothing drew before

| panel | what it is for |
|---|---|
| `plot_corner` | pairwise posterior. Where a fit is unidentifiable this is where it shows — a diagonal ridge means two parameters trade off and only their combination is determined, which a marginal histogram hides completely. |
| `plot_posterior_predictive` | model draws against the measurements, everything in units of **each measurement's own std**, so 84 observables on different scales share one axis and "inside the error bar" means the same distance everywhere. |
| `plot_coverage` | coverage against its nominal level. Bars are labelled directly, because a coverage of 0 is a bar of no height and an unlabelled empty slot reads as "not measured" rather than "none of them". |

All three read what `libcuflynx.param_id.posterior_predictive` writes and return quietly when it is absent, so they are safe to leave on for a calibration-only run.

## Three bugs found by drawing them, not by reading them

Against a real 84-observable, 18-parameter run:

- **`tex()` wraps a label in `$…$`, and a parameter called `V_mid_m_Kv4_2` is not maths.** Mathtext reads the second underscore as a second subscript and raises `Double subscript`, losing the whole figure over one axis label. New `plain()` leaves arbitrary identifiers outside maths mode, where matplotlib renders them literally.
- **`paginate()` returns pages of items, not `(start, end)` pairs.** My first version unpacked it and lost the panel to `too many values to unpack`.
- **`save()` did not crop to contents**, so a legend below the axes and a long axis label were cropped at the figure edge — silently, since the file is still written. Now `bbox_inches="tight"`, which helps every existing panel too.

## Checked by drawing

Generated a bundle for a real study and ran its `plot_outputs.py` against a finished run: `posterior_corner.png`, `posterior_coverage.png`, and three pages of `posterior_predictive_*.png`, no warnings. The coverage panel reported 0% / 8% at 80% and 0% / 14% at 95% — over-confident and biased, which the predictive panel then shows as observables sitting up to 65 std from the data. That is the check earning its place on its first run.

29 tests.


FinbarArgus:
Closed and immediately reopened to get a check run attached — no change to the branch.

This PR was retargeted to `master` after it was opened. The Tests workflow filters on `pull_request: branches: [master, main, devel]`, and a base change fires `pull_request: edited`, which the workflow's default event types do not include — so it had never had a PR-attached run despite being mergeable. A `workflow_dispatch` run does not attach to a PR either, which is why the manual run earlier did not show up here.

The branch is up to date with `master`, so it now carries the `Cython<3.3` build constraint from #472 that the earlier dispatch failed on.

Issue #466, opened 2026-08-20
Test on accessing predictions from different (sub)-experiments


MOH-Shafizadegan:
The current test using `3compartment_extra_ops_obs_data.json` only assess the `calculate_two_observable_difference` functionality within a single (experiment, subexperiment). This doesn't test the ability to access predictions/variables from different experiments and subexperiments (#127 ).

Having a test to check such functionality between different experiments and sub-experiments is recommended.

FinbarArgus:
Agreed, good issue

PR #482, opened 2026-08-23
release: libcuflynx 0.5.0

FinbarArgus:
Cuts **libcuflynx 0.5.0**. Built on #481 (which is fully green — all 11 jobs), so this PR is the
release commit on top of #478, #479 and #480.

Version bump and changelog only. No behaviour changes.

## Releasing

Merge, then tag `v0.5.0`. `release.yml` fires on `v*`, checks the tag against `pyproject.toml`'s
version — PyPI never accepts a version twice — and publishes by trusted publishing.

## Why 0.5.0 and not 0.4.2

This release carries a breaking change: #466 separates an obs_data item's name from its labels,
and the CHANGELOG says so in as many words. A patch number would understate that for anyone
reading the version alone.

## The flat-import shims survive it, and now go in 0.6.0

They were promised for removal **in** 0.5.0 — `REMOVAL_VERSION`, eleven shim docstrings, the
README, CONTRIBUTING, CLAUDE.md and the tutorial all said so.

Removing them here would make one upgrade demand **two unrelated migrations**: one touching
users' obs_data, one touching their imports. So the promise moves to 0.6.0 rather than being
quietly broken, and every place that stated it is updated in step.

**The load-bearing part:** three test files restated the literal `"0.5.0"` rather than reading the
constant — `test_docs_consistency`, `test_deprecated_flat_imports`, `test_release_workflow`. They
now read `REMOVAL_VERSION`. Left as literals they would have passed while every document said
something else, which is precisely the drift those tests exist to catch.

CHANGELOG's **0.4.0** entry still reads "go away in 0.5.0". Deliberate: it records what that
release announced, not a claim about today. The deferral is recorded in the new 0.5.0 entry.

## Verified

- CA unit tier: **1429 passed, 0 failed**
- `python -m build` produces `libcuflynx-0.5.0-py3-none-any.whl` and `libcuflynx-0.5.0.tar.gz`
- the three doc/release guard suites: 51 passed

## Note on CI

Based on `integration/next`, so **no workflow runs on this PR** — the same property that hid
#470's defect last round. Two mitigations: everything here is covered by the local run above, and
GitHub retargets this to `master` automatically once #481 merges, at which point CI does run. Worth
letting it go green then, before tagging.

## Testing CUFLynx against it before the release

You do not need to wait for PyPI. The wheel is built and installed into
`~/.claude/venvs/cuflynx-ci`, so CUFLynx can be pointed at that interpreter now — see the comment
below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ next release; add to that section as you land a change.
 
 ## Unreleased
 
+## 0.5.0 — 2026-08-24
+
+### Changed — the flat-import shims now go away in 0.6.0, not 0.5.0
+
+They were promised for removal in this release. They survive it instead. This release already
+asks users to migrate their obs_data files (below), and removing the `libcuflynx.` namespace
+shims at the same time would make one upgrade demand two unrelated migrations — one touching
+data, one touching imports. `REMOVAL_VERSION` now reads `0.6.0`, and every shim's warning, the
+README, CONTRIBUTING, CLAUDE.md and the tutorial say so.
+
+The tests that check those docs agree now read `REMOVAL_VERSION` rather than restating it: the
+literal `"0.5.0"` in three test files would have passed happily while every document said
+something else.
+
+
 ### Changed — the scaling benchmark now reports how much work it did (#344)
 
 The 3compartment core-scaling sweep reported CMA-ES speeding up **22.8x on 8 cores**, past the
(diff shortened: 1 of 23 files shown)


FinbarArgus:
## Testing CUFLynx against 0.5.0 now, without waiting for PyPI

The wheel this PR builds is installed into `~/.claude/venvs/cuflynx-ci`, as a real
distribution in `site-packages` rather than a source checkout:

```
libcuflynx 0.5.0
  REMOVAL_VERSION: 0.6.0
  DEFAULT_COST_TYPE: gaussian_MLE
  autoemulate: 2.1.2      (installed with --build-constraint Cython<3.3, per #472)
```

To test CUFLynx against it:

1. **Settings → Interpreter** (top bar) → `/home/farg967/.claude/venvs/cuflynx-ci/bin/python`.
   This is what calibration / sensitivity / UQ actually run in.
2. **Settings → CA dir** → clear it. That is the step people miss: a CA dir sets
   `CIRCULATORY_AUTOGEN_SRC`, which wins over the installed package for the in-process engine,
   so leaving it pointed at a checkout means the live plots and the obs-data editor are still
   reading source while the analyses read 0.5.0 — two different engines in one session.

Then restart the app: CA modules are cached after the first simulation, so re-pointing mid-session
only fully takes effect on a restart.

To rebuild the wheel after any change here:

```bash
python -m build --outdir dist .
~/.claude/venvs/cuflynx-ci/bin/pip install --force-reinstall --no-deps dist/libcuflynx-0.5.0-py3-none-any.whl
```

Issue #468, opened 2026-08-21
Bug: joblib.dump fails  when ABC metadata (_abc._abc_data) evaluates to None


MOH-Shafizadegan:
When saving an emulator bundle using `joblib.dump(model, ...)`, serialization crashes if components like model.scheduler_setup or model.to contain or reference abstract base class implementation data (`_abc._abc_data`) that resolves to `None`.

Standard `joblib` and Python's pickle fail to handle this uninitialized or None-valued C-extension descriptor state, resulting in a pickling crash:
```bash
FAILED on attribute: model.scheduler_setup -> cannot pickle '_abc._abc_data' object
FAILED on attribute: model.to -> cannot pickle '_abc._abc_data' object
...
TypeError: 'NoneType' object is not iterable (inside dill/pickle post-processing)
```

### Proposed Fix:
Use dill instead of joblib for serializing the model in `emulator_bundle.py`, as dill robustly handles these internal ABC implementation states and C-extension references:
```python
def _save_model(model, path_without_suffix):
    """dill, the same container autoemulate's own serialiser uses."""
    
    import dill
    with open(f'{path_without_suffix}.joblib', 'wb') as f:
        dill.dump(model, f)
```

FinbarArgus:
@MOH-Shafizadegan could you please send me a simple model and run script that replicates this? Do you know why this model has abc data?

FinbarArgus:
Tried to reproduce this against a real study — the `control_parasymp` model from `CA_user_FA_MG` (23 parameters, 36 scalar features, 6 experiments) — and **could not make `joblib.dump` fail**, on any autoemulate version in the supported range.

### What was run

A 12-point design, trained through `cuflynx-train-emulator` across 12 MPI ranks (one sample each; a single sample of this model is ~27 min of CPU, so this is the cheapest design that autoemulate will fit at all — 3 points gives `ValueError: No results available`). It completed and saved:

```
OK in 1534.0s   emulator.joblib 1.7 MB   model_name: GaussianProcessRBF
```

### Every model class, on that same design

`emulator_settings.reuse_samples: true` refits the saved `training_data.npz` without re-simulating, so all twelve of `emulator_model_names()` were tried:

| result | models |
|---|---|
[...]

PR #497, opened 2026-08-25
Let the emulator choose how it is pickled, and fall back rather than lose a run (#468)

FinbarArgus:
Closes the practical half of #468.

## The cost of the bug

Training pays for **every simulation before it saves**. So a fitted model joblib cannot pickle does not cost a save, it costs the whole run — which is what makes #468 expensive rather than annoying:

```
TypeError: cannot pickle '_abc._abc_data' object
```

## What changed

`emulator_settings.model_serialiser` — `auto` (default), `joblib`, `cloudpickle`, `dill`.

`auto` works down joblib → cloudpickle → dill until one succeeds, removes the half-written file on the way, and prints which container it ended up using. Which one wrote a bundle is recorded in `emulator_metadata.json`; load prefers the recorded one and tries the rest, so bundles written before this still open.

## Why a chain, and not the switch to dill the issue proposes

Because dill is not a superset of joblib. Measured on autoemulate 2.1.2, fitting real emulators and round-tripping them:

| container | an object pickle cannot name | a torch-backed emulator |
|---|---|---|
| joblib | fails | works |
| cloudpickle | works | works |
| **dill** | works | **fails** — recurses on a `PyCapsule` |

Adopting dill outright would have broken the common case to fix the rare one. joblib therefore stays first (it is what autoemulate itself writes and reads), and every real emulator tried — GP-RBF, MLP, random forest — still saves with joblib under `auto` and reloads predicting identically.

cloudpickle is second because it was the only one that handled both columns.

## On not being able to reproduce it

The `_abc._abc_data` object is autoemulate's and I could not produce one on demand, so the tests reproduce the *mechanism*: an object pickle refuses and the others accept. That paid for itself immediately — the first version of the fallback did not fire, because the C pickler raises `TypeError` (what the issue quotes) while the Python one raises `PicklingError`. Catching only the quoted exception would have shipped a fallback that missed half the cases.

What remains unverified is whether the reporter's specific emulator is one cloudpickle or dill can take apart. The mechanism is right and both handle strictly more than joblib in the failing direction, but if it fails for another reason it will still fail — the error now names the setting and the containers tried, so we would hear how.

## Also

- `cloudpickle` and `dill` are declared in the `emulation` extra rather than relied on. Both arrive transitively today (`multiprocess`, `schwimmbad`, joblib's vendored copy) and that is not a promise.
- CUFLynx needs no change: its emulator settings form is generated from `ANALYSIS_OPTIONS`, so the dropdown appears on its own.

## Testing

`tests/test_emulator_serialiser.py` — 18 tests: the fallback, the ordering and why, no half-written file, each container round-tripping what joblib refuses, the recorded choice, and bundles that predate the setting. Plus the emulator training suite (15) and the schema guard.

## For the reporter

    pip install "libcuflynx[emulation] @ git+https://github.com/physiomelinks/circulatory_autogen@fix/468-emulator-serialiser"

Then train as usual. `auto` needs no configuration; if it still fails, the message will name what was tried.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -185,6 +185,16 @@ uq = [
 # only when do_emulation/use_emulator is set, and says so by name when it is missing.
 emulation = [
     "autoemulate>=2.1,<3",
+    # Fallback containers for saving a fitted emulator. Some models hold an
+    # uninitialised C-extension descriptor that pickle cannot take apart, and
+    # joblib.dump then fails after every training simulation has been paid for
+    # (issue #468). Neither is a superset of joblib -- dill fails on torch-backed
+    # emulators joblib saves fine -- which is why they are fallbacks in order
+    # rather than a replacement. Declared rather than relied on: both arrive
+    # transitively today and that is not a promise.
+    # See emulator_settings.model_serialiser.
+    "cloudpickle>=2.0",
+    "dill>=0.3.6",
 ]
 # `model_type: cpp` (and the 0D-1D coupling built on it) generates C++ source for the user to
 # compile. What it needs is a toolchain -- a C++ compiler, CMake, and for the 1D arm PETSc --
--- a/src/libcuflynx/emulators/emulator_bundle.py
+++ b/src/libcuflynx/emulators/emulator_bundle.py
@@ -16,11 +16,35 @@ metadata -- an emulator reloaded without them would predict in the wrong units.
 import hashlib
 import json
 import os
+import pickle
 
 import numpy as np
 
 METADATA_FILE = 'emulator_metadata.json'
 MODEL_FILE = 'emulator'          # autoemulate's saver appends .joblib
+#: How the fitted model is pickled. The file is called ``emulator.joblib``
+#: whichever is used -- that is the artefact's name, not a claim about the
+#: container -- and the bundle records which one wrote it so it can be read back.
+SERIALISERS = ('auto', 'joblib', 'cloudpickle', 'dill')
+DEFAULT_SERIALISER = 'auto'
+#: What ``auto`` tries, in order. joblib first because it is what autoemulate
+#: itself writes and reads. cloudpickle before dill on measurement, not
+#: reputation: on autoemulate 2.1.2 dill *fails* where joblib succeeds --
+#: torch-backed emulators hold a ``PyCapsule`` it recurses on -- so the fix
+#: proposed in #468 would have traded one broken case for a commoner one.
+#: cloudpickle handled every emulator tried, and the unnameable objects joblib
+#: refuses. dill stays last because it is what the issue reports working.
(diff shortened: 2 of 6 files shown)


FinbarArgus:
Waiting on check from @MOH-Shafizadegan to see if this fixes his issue.

FinbarArgus:
## Second failure from the same reporter, and one this PR's fallback chain cannot fix

```
PicklingError: Can't pickle sentinel: it's not the same object as typing_extensions.sentinel
```

Reproduced exactly, in one line, and it is **not** about which container is used.

### What it is

A [PEP 661](https://peps.python.org/pep-0661/) sentinel pickles *by name*: `__reduce__` returns a string, pickle stores it as a global, and on the way back in checks that the name still refers to the same object. `typing_extensions` 4.16.0 ships one where that check cannot pass — line 273:

```python
_marker = sentinel("sentinel")     # named "sentinel", bound to _marker
```

`typing_extensions.sentinel` is the *class*, so anything holding `_marker` cannot be pickled by anything:

| container | `typing_extensions._marker` |
|---|---|
| joblib | `PicklingError` |
| cloudpickle | `PicklingError` |
| dill | `PicklingError` |

All three defer to the object's own `__reduce__`, so there is nothing to fall back **to**. Choosing a different container was the right answer for #468's `_abc._abc_data`; this needed a different one.

### Version, and why CI never saw it

| typing_extensions | `_marker` | pickles? |
|---|---|---|
| 4.15.0 | `_Sentinel` instance, default `__reduce__` | fine |
| **4.16.0** | `sentinel("sentinel")`, `__reduce__ → "sentinel"` | **broken** |

The bug arrived in 4.16.0, which is the current release — there is no newer version to upgrade to. This repo's own venv is on 4.15.0, which is why nothing here ever hit it.

### The fix

Look the object up by identity in its own module and pickle it as the name that *does* resolve (`_marker`, not `sentinel`), registered through `copyreg.dispatch_table` — which every pickler built on `pickle.Pickler` consults **before** `__reduce_ex__`, so one registration covers all three containers. Scoped to the dump and removed afterwards, because that table is process-wide; an existing registration for the same class is left alone.

Identity is preserved, not merely picklability. A sentinel is a singleton whose entire purpose is identity, so returning an equal copy would be a silent behaviour change dressed up as a fix.

### Testing

The tests skip on **behaviour** rather than version — they ask whether the installed `typing_extensions` can pickle its own sentinel — so if upstream fixes this they start skipping instead of passing vacuously.

- under 4.16.0: **32 passed**
- under 4.15.0: 21 passed, 11 skipped
- emulator training + solver helper suites: 36 passed

### For the reporter

Re-install the same branch. There is also a workaround that needs no branch at all:

    pip install "typing_extensions!=4.16.0"

Issue #331, opened 2026-07-28
Support calibrating total volume (not initial LV volume) so offline_pre_time can be used


FinbarArgus:
## Why

Calibration now **refuses** `offline_pre_time` when any calibration parameter sets a state's initial value, because an offline warm-up freezes one state and restarts every evaluation from it — so such a parameter never enters the simulation. Measured on `3compartment`: `q_lv_init`'s response to a +25% perturbation drops from **8.49% to exactly 0.0000%** once an offline warm-up is used, and its gradient becomes exactly zero. That is worse under gradient-based optimisation than under a GA, since the optimiser sees a flat direction and the parameter is silently unrecoverable.

The error tells the user the way out:

> varying initial state (quantity) requires doing it from the actual initial state, so offline_pre_time can't be used. Reformulate to calibrate wrt a constant parameter rather than a initial state if offline_pre_time is essential. e.g. rather than setting initial LV volume, set the total volume and calculate the intial LV volume from that.

This issue is to make that reformulation actually possible.

## What is needed

A model/parameter formulation where the calibrated quantity is a **constant** that the initial states are derived *from*, rather than an initial state itself. For the circulatory models the natural choice is **total blood volume**:

- calibrate `V_total` (a constant),
- distribute it across compartments at initialisation — `q_lv_init = f(V_total, ...)`, likewise for the other compartment volumes.

Because `V_total` is a constant rather than a state initial value, it keeps affecting the trajectory when the run starts from a frozen offline state (the offline pass would need the distribution recomputed at the trial `V_total`, which is cheap and parameter-local), so `offline_pre_time` becomes usable for gradient-based calibration of the volume parameter.

## Considerations

- The distribution rule must be part of the model (CellML), so the generated code and its derivatives see it — not applied externally, or AD/FSA will miss `d(initial state)/d(V_total)`.
- Conservation: in a closed loop total volume is conserved, which is exactly why the current `q_lv_init` formulation is so sensitive; the reformulation should make that conservation explicit rather than implicit in one compartment's initial condition.
- Once available, `3compartment` (and the benchmark) could switch from `q_lv_init` to `V_total` and then benefit from `offline_pre_time`.
- Interacts with #328 (how long the per-evaluation warm-up needs to be) and #269 (correct offline warm-up reuse).

Issue #332, opened 2026-07-28
Modify model parameters/formulation to speed up convergence to periodic steady state


FinbarArgus:
## Why

Warm-up dominates calibration cost. On the `3compartment` benchmark, `pre_time: 20` against `sim_time: 2` means **91% of every integration is discarded warm-up**, and integration is ~94% of total calibration runtime (measured, see `benchmarks/OPTIMISATION_OPPORTUNITIES.md`). Anything that makes the model reach periodic steady state sooner is close to a proportional saving on the whole calibration.

## How slow is it

Time for `aortic_root/u` to settle within 1% of its final cycle:

| trial parameters | from cold initial conditions | from a converged baseline state |
|---|---|---|
| baseline | 6 s | 0 s |
| box corner (low) | 24 s | 8 s |
| box corner (high) | 22 s | 15 s |
| mid-box | — | 9 s |

Two observations:

1. Cold convergence is slow enough (up to 24 s, i.e. ~24 cardiac cycles) that the current `pre_time: 20` is **not sufficient at the low corner** — those evaluations are scored before the model has settled.
2. Even warm-started, the corners still need 8-15 s, so the settling rate itself is the limiting factor.

## Possible directions

- **Better initial conditions.** The CSV initial values are far from any operating point (`q_lv_init` default 0.002 m^3 is outside the identification bounds of [2e-4, 1.5e-3]). Initialising compartments near a physiological distribution should cut the cold transient substantially.
- **Reduce the slowest time constant.** Identify which compartment/parameter governs the slow mode (venous compliance is the usual suspect in lumped circulatory models) and check whether its value is physiologically justified or merely inherited.
- **Initialise from an analytic/steady-state solve** rather than integrating the transient away — solve for the periodic orbit directly, or at least for the mean operating point, and start there.
- **Shorter cycle-count requirement.** If the observables are cycle-averaged quantities, some may converge in far fewer cycles than the slowest state; warming up only until the *observed* quantities settle (rather than all states) may be enough.

## Related

- #328 — detect periodic steady state during warm-up instead of using a fixed `pre_time`; that adapts the duration, whereas this issue is about making the duration needed smaller in the first place.
- #331 — reformulating volume parameters, which also affects how the initial state is set.
- #269 — reusing an offline warm-up across evaluations.

PR #507, opened 2026-08-31
One spelling per concept in obs_info and prediction_info

FinbarArgus:
One spelling per concept in `obs_info` and `prediction_info`.

## The problem

#466 split the obs_data **file** vocabulary — `variable` → `data_item_name` + `operands`, `name_for_plotting` → `trace_name_for_plotting` + `item_name_for_plotting` — and the file layer got a proper rule: rewrite on load, warn once, refuse a document setting both spellings. The **in-memory** dicts never followed, leaving three inconsistencies:

1. **The parser wrote the old name as the real one.** `obs_info["obs_names"]` was the list, with `data_item_names` aliased to the same object — so the key a user writes and the key the engine holds disagreed for no reason.
2. **`names_for_plotting` meant two different things** — the *item* label in `obs_info`, the *trace* label in `prediction_info`. One key, two meanings, and `prediction_info` had no accessors to hide it.
3. **Ten sites read the deprecated spellings directly**, bypassing the accessors that exist to hide them. Two reimplemented the fallback chain with a *different* precedence.

## The shape

`obs_info` now has `data_item_names` / `trace_names_for_plotting` / `item_names_for_plotting` and nothing else. `prediction_info` mirrors it exactly: `names` → `operands` (a list per item, as on obs_info), `names_for_plotting` → `trace_names_for_plotting`. `obs_item_names` / `obs_item_labels` / `obs_trace_labels` / `obs_operand_lists` work on either dict.

A dict handed in from outside is normalised at `ParamID.__init__` and the two setters — warning once per superseded key, erroring if it sets both an old key and its replacement. That is the same bargain `migrate_legacy_obs_item_keys` already strikes for files; the accessors no longer absorb old names silently, because a silent fallback told nobody anything and kept the old spelling alive indefinitely.

**Two legacy tables, deliberately not one.** `names_for_plotting` maps to *different* targets per dict, so a shared mapping would move trace labels into the item-label slot (or the reverse) and nothing would fail — an axis would just be captioned wrongly. `test_the_two_info_tables_disagree_on_purpose` exists to be read before somebody merges them.

## Seven commits, green at each

`b160ad77` reads → accessors · `1f496ab7` tables + normalisers · `ecdedc6b` parser canonical only · `20b1cc18` normaliser at the boundary · `b947c376` accessor fallbacks stripped · `0fb86622` docs + the guard · `e37ce615` the npz key set

## Bugs found on the way

- **`posterior_predictive` captioned a series panel with the item's identity** where it meant the trace label — the fallback chain tried identity first.
- **`print_observable_errors` named one of six branches by identity** where the other five used the label.
- **`test_no_label_restates_its_own_operation` could not fail.** It read `name_for_plotting`, which the migrator renames on load, so it was always handed `None`. Repointing it alone would have been a smaller version of the same mistake — no shipped file sets `item_name_for_plotting` on a reducing item either — so `_spells_its_own_operation` is now tested directly. A clean corpus and a broken check are indistinguishable from outside, which is how this rotted for a release.
- **The tutorial documented the old vocabulary for `prediction_items`** while the schema required the new one; the `data_items` section 300 lines above was updated at the time and this was missed.
- **`all_outputs_with_best_param_vals_exp_0.npz` was not reproducible** — see below.

## The npz (`e37ce615`)

Across five runs of one study, identical model and identical code, the file came back with 76 keys four times and 77 the fifth. The extra was `volume_sum_module.t`, holding an array bit-identical to `environment.time`.

The model connects them, and Myokit merges connected variables on import, keeping one qname for the pair:

```xml
<map_components component_1="environment" component_2="volume_sum_module"/>
<map_variables variable_1="time" variable_2="t"/>
```

`environment.time` is not in the model at all — CA synthesises it as a stable name — and the log arrives keyed both ways. CA appended its stable key *alongside* whatever the log had rather than instead of it. Fixing the key set alone was not enough: the time's position still tracked the log's shape, so it is now written once and first.

## Verification

A full pipeline run was captured before any of this and diffed at every commit. **All 22 artifacts byte-identical**, including `error_vec_names.npy` (a documented external contract) and `prediction_variable_names.csv` (which proves the `operands` reshape moved no qname). The npz was excluded mid-branch because it could not be trusted, and is back under the contract as of `e37ce615` — two consecutive runs identical, and identical to the pre-branch baseline.

Suite: **2215 passed, 4 failed** — the AADC/CasADi four, verified to fail identically on clean `master`. Also run under `-W error::DeprecationWarning` to prove the normaliser is silent on every parser-produced dict.

## Downstream

Paired with **physiomelinks/CUFLynx#331**. CUFLynx is not broken in the meantime — its three reads are `or`-fallbacks whose first arm is `item_names_for_plotting`, which the parser always populates. Merge this, bump the two `ref:` lines in CUFLynx's `ci.yml`, then merge that.

https://claude.ai/code/session_01Mk1rvBawTUN3SP7VwHVGhh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,29 @@ next release; add to that section as you land a change.
 
 ## Unreleased
 
+### Changed — one spelling per concept in `obs_info` and `prediction_info`
+
+The obs_data *file* vocabulary was split by #466; the parsed dicts were not, and kept the old
+spellings as live aliases. `obs_info` held `obs_names` with `data_item_names` aliased to the
+same list, and `names_for_plotting` aliased to `item_names_for_plotting`. `prediction_info`
+was worse: its `names_for_plotting` meant the **trace** label where obs_info's meant the
+**item** label, so one key named two different things depending on which dict you held.
+
+Now: `obs_info` has `data_item_names`, `trace_names_for_plotting` and
+`item_names_for_plotting` and nothing else; `prediction_info` mirrors it exactly, with its
+`names` becoming `operands` (a list per item, as on obs_info) and its `names_for_plotting`
+becoming `trace_names_for_plotting`. `obs_item_names` / `obs_item_labels` /
+`obs_trace_labels` / `obs_operand_lists` now work on either dict.
+
+**If you build one of these dicts by hand**, `ParamID` normalises what it is handed and warns
+once per superseded key; a dict setting both an old key and its replacement is an error.
+Reading the old keys off a parsed dict no longer works, because they are no longer written.
+
+Two bugs fell out. `posterior_predictive` captioned a series panel with the item's *identity*
+where it meant the trace label, and `print_observable_errors` named one of its six branches by
+identity where the other five used the label.
+
+
 ## 0.6.0 — 2026-08-31
 
 ### Removed — the flat-import shims (#428)
@@ -31,6 +54,7 @@ internally (`paramID.py`), so removing it is a change to the parser and its call
 than a deletion, and it does not belong in the same release as the namespace removal for the
 same reason that removal was deferred out of 0.5.0. Prefer `data_item_names`,
 `trace_names_for_plotting` or `item_names_for_plotting`; the alias will go in a later release.
+*(Done in Unreleased — see "one spelling per concept" above.)*
 
 ## 0.5.1 — 2026-08-25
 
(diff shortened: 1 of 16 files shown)


FinbarArgus:
Merging with `test-cuflynx-executable` red, deliberately.

That job downloads the latest CUFLynx **release** (v0.4.1) and points it at the CA under test. v0.4.1 predates the obs_data vocabulary split and reads two keys with no guard — `local_sensitivity.py:340` and `obs_cost.py:442` — so it raises `KeyError: 'names_for_plotting'` against this branch, and 500s on `/api/simulate`.

The fix has been on CUFLynx `main` since `bbf16fd` and is now also in physiomelinks/CUFLynx#331; it has simply never been in a release. Cutting CUFLynx 0.5.0 is what turns this green, and that release is being prepared in physiomelinks/CUFLynx#332 so it can bundle libcuflynx 0.7.0 rather than 0.6.0.

The job is not a required check. Nothing else on this PR is red.

